PR #104, opened 2026-04-10
feat(comark): add footnotes plugin with parsing and rendering support

### LINKED ISSUE #103: [Feature Request]: footnotes support
# Description

Standard markdown footnotes ([^1]) are not currently handled, it would be nice to have the support

example: https://github.blog/changelog/2021-09-30-footnotes-now-supported-in-markdown-fields/

- [x] I'd be happy to work on it

arashsheyda:
closes #103

### How to use

run the example and check it out for yourself

example text:

```
Here is a simple footnote[^1]. With some additional text after it.

[^1]: My reference.
```

Screenshot 2026-04-10 at 2 35 32 AM


packages/comark/src/plugins/footnotes.ts
```diff
@@ -0,0 +1,189 @@
...
+          }
+
+          // Recurse into children of element nodes
+          const [tag, attrs, ...children] = node as ComarkElement
+          const processedChildren = processNodes(children as ComarkNode[])
+          result.push([tag, attrs, ...processedChildren] as ComarkNode)
+        }
+
+        return result
+      }
```
farnabaz: I believe it will be easier with visit utility

```suggestion
visit(
        state.tree,
        (node) => Boolean(isFootnoteRef(node)),
        (node) => {
          const refLabel = isFootnoteRef(node)
          if (refLabel && definitions.has(refLabel)) {
            if (!refIndexMap.has(refLabel)) {
              refIndexMap.set(refLabel, refIndexMap.size + 1)
            }
            const refIndex = refIndexMap.get(refLabel)!

            return ['sup', { class: 'footnote-ref' },
              ['a', {
                href: `#fn-${refLabel}`,
                id: `fnref-${refLabel}`,
              }, `[${refIndex}]`],
            ]
          }
        },
      )
```

farnabaz:
Did you try the official ‎`markdown-it` footnote plugin? https://github.com/markdown-it/markdown-it-footnote

We could compare the performance and usability of the two implementations.

I prefer implementing footnotes within our transformer because it gives us more flexibility, but it’s worth checking the official plugin and comparing.

arashsheyda:
> Did you try the official ‎markdown-it footnote plugin? [markdown-it/markdown-it-footnote](https://github.com/markdown-it/markdown-it-footnote)

oh no I wasn't aware of this, will take a look!

packages/comark/src/plugins/footnotes.ts
arashsheyda: updated

arashsheyda:
@farnabaz so I asked copilet to do a benchmark and here's the result:

Screenshot 2026-04-13 at 7 27 14 PM


and then later I asked to improve it and compare the code in [markdown-it-footnote](https://github.com/markdown-it/markdown-it-footnote) and pushed the changes and here's the new improved benchmark:

Screenshot 2026-04-13 at 7 28 40 PM

(credits to copilet)

packages/comark/src/plugins/footnotes.ts
```diff
@@ -0,0 +1,189 @@
...
+ */
+export default defineComarkPlugin((config: FootnotesConfig = {}) => {
+  const {
+    label = 'Footnotes',
+    hr = true,
+    backRef = '↩',
+  } = config
+
+  // Store definitions extracted during `pre` for use in `post`
+  let definitions: Map<string, string>
```
farnabaz: Lets move this into state.context to prevent mixing definitions of two  parallel parse

farnabaz:
Just notices about the definitions, I think it will cause issue in parallel parses.
Other than that, everything looks good 👍 


Thanks for the benchmark, lets move forward with current implementation 👍

packages/comark/src/plugins/footnotes.ts
arashsheyda: updated, I might've not understood fully so let me know if I did the right thing. thanks

farnabaz:
Looks good to me 👍 

I just drop a PR #130 to allow supporting conditional node rendering. Lets wait for that PR and then we can implement footnote stringify component so users can have valid markdown when they want to render markdown from Comark Tree

Heres is simple handler/component for stringify:
```ts
{
  match: (node) => {
    return node[1].class === 'footnotes' || node[1].class === 'footnote-ref'
  },
  handler: (node, children, options) => {
    if (node[1].class === 'footnotes') {
      const ol = node.find(n => n[0] === 'ol') as ComarkElement
      let result = ''
      for (let i = 2; i < ol.length; i++) {
        const key = String((ol[i][1] as ComarkElementAttributes)?.id)?.replace('fn-', '')
        const value = ol[i][2] as string
        result += `[^${key}]: ${value}\n`
      }
      return result
    }
    if (node[1].class === 'footnote-ref') {
      const link = node[2] as ComarkElement
      const key = String((link[1] as ComarkElementAttributes)?.id)?.replace('fnref-', '')
      return `[^${key}]`
    }
    return ''
  },
}
```

packages/comark/src/plugins/footnotes.ts
```diff
@@ -0,0 +1,190 @@
...
+      const definitions = new Map<string, string>()
+
+      // Extract and remove footnote definitions from the source
+      state.markdown = state.markdown.replace(FOOTNOTE_DEF_RE, (_match, defLabel: string, content: string) => {
+        definitions.set(defLabel, content.trim())
+        return '' // Remove the definition line
+      })
+
+      // Store on state to avoid mixing definitions across parallel parses
+      state.context.footnote = definitions
```
farnabaz:
```suggestion
      state.footnote = definitions
```

packages/comark/src/plugins/footnotes.ts
```diff
@@ -0,0 +1,190 @@
...
+    },
+    // replace [^ref] spans and build footnotes section
+    post(state) {
+      const definitions: Map<string, string> = state.context.footnote
```
farnabaz:
```suggestion
      const definitions: Map<string, string> = state.footnote
```

arashsheyda:
@farnabaz added a `footnotesStringify`, could you take a look and verify :ty:

packages/comark/src/plugins/footnotes.ts
```diff
@@ -0,0 +1,236 @@
...
+ * const tree = await parse('Hello[^1]\n\n[^1]: World', {
+ *   plugins: [footnotes()]
+ * })
+ *
+ * const md = await renderMarkdown(tree, {
+ *   components: { footnotes: footnotesStringify },
+ * })
+ * ```
+ */
+export const footnotesStringify: ConditionalNodeHandler = {
```
farnabaz: To be consistent with all plugins and since this is just a handler, lets call it `Footnote`

```suggestion
export const Footnote: ConditionalNodeHandler = {
```

arashsheyda: updated

farnabaz (approved):
LGTM 👍 
Thanks

PR #106, opened 2026-04-10
fix(specs): preserve indentation of nested components in named slots

larbish:
Fix renderMarkdown stripping the indentation of nested components at the start of named slots, breaking the parse/serialize round-trip.

`.trim()` was removing the leading 2-space indent from the first child of a slot's rendered content.
Replacing it with `.trimEnd()` preserves leading whitespace while still stripping trailing newlines.

Two specs covering the affected scenarios (`component-nested-unindented.md`, `component-slot-nested-components.md`)


farnabaz (approved):
👍

PR #109, opened 2026-04-11
fix(examples): add tailwindcss to dependencies

arashsheyda:
# Description

when running these 2 examples, user would get an error: `Can't resolve 'tailwindcss' in `

this fixes it


farnabaz (approved):
Thanks

PR #116, opened 2026-04-14
chore: add pnpm catalog support

### LINKED ISSUE #110: [Suggestion]: pnpm catalog
# Description

I noticed we’re repeating the same dependencies across different examples and packages. it might be easier to manage if we move them into a [pnpm catalog](https://pnpm.io/catalogs) so versions stay consistent and updates are simpler across the repo.

happy to take this on if it sounds good

arashsheyda:
# Description

closes #110 

moved all dependency versions to a centralized catalog in `pnpm-workspace.yaml`, all package.json files now reference versions via `catalog:` protocol instead of inline ranges

wider peer dep ranges in published packages (e.g. katex >=0.16, beautiful-mermaid >=1.0) are kept explicit since the catalog can only hold one version per package name

@farnabaz we can also use categorized catalogs like we do in devtools https://github.com/nuxt/devtools/blob/main/pnpm-workspace.yaml or keep it like this. what do you think?

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -5,25 +5,25 @@
     "build": " pnpm --filter '../' run build && nuxt build --extends docus"
   },
   "dependencies": {
-    "@comark/nuxt": "^0.2.0",
-    "@monaco-editor/loader": "^1.7.0",
-    "@nuxt/content": "^3.12.0",
-    "@nuxt/ui": "^4.6.1",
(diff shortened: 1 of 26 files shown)


farnabaz:
I think this is good start, let keep single catalog for now and we create separate ones if it needed

farnabaz (approved):

PR #120, opened 2026-04-14
fix(render): correct list indentation and block content in list items

### LINKED ISSUE #114: renderMarkdown nested lists indentation is off inside ordered lists
Hi! First of all, thanks for comark — really nice project, the MDC-aware parsing is exactly what we needed 🙏

I'm using `parse()` + `renderMarkdown()` as a markdown formatter (roundtrip), and I noticed that nested unordered lists inside ordered list items loose their indentation level.

**Input:**
```markdown
1. **First item** — Description
   - sub-item a
   - sub-item b

2. **Second item** — Description
   - sub-item c
```

**Output after roundtrip:**
```markdown
1. **First item** — Description
  - sub-item a
  - sub-item b
2. **Second item** — Description
  - sub-item c
```

The sub-items get 2 spaces indent instead of 3. According to CommonMark spec, the continuation content should be indented to match the list marker width (`1. ` = 3 characters). With only 2 spaces, a CommonMark parser will not reconize the sub-items as nested anymore, wich breaks the structure on a second pass.

I think the issue comes from `ol.js` — the context don't carry the prefix width, so when a nested `ul` calls `indent(result)`, it always default to 2 spaces (`'  '.repeat(1)`), even when the parent is an ordered list.

A possible fix would be to track `listIndent` in the context (3 for `ol`, 2 for `ul`) and use it in the `indent()` call for nested lists. I have a working patch if that helps!

### LINKED ISSUE #118: renderMarkdown merges code blocks into preceding text inside list items
When a list item contains both inline text and a fenced code block, `renderMarkdown` concatenates them without a newline separator and loses the indentation required by the list context.

### Reproduction

Input:

````markdown
- Item with code:
  ```bash
  echo "hello"
  ```
- Next item
````

Output after `renderMarkdown`:

````markdown
- Item with code:```bash
echo "hello"
```
- Next item
````

Expected output:

````markdown
- Item with code:
  ```bash
  echo "hello"
  ```
- Next item
````

### Root cause

In `dist/internal/stringify/handlers/li.js`, the handler iterates over all children and concatenates them without distinguishing inline content (strings) from block content (`pre`, `ul`, `ol`, `blockquote`). The `result.trim()` on line 17 further strips the newlines that `pre` handler adds.

The `pre` handler returns its content with `blockSeparator` appended, but since `li` trims the result, the newline before the code fence is lost and no indentation is applied to align the block with the list prefix.

### Suggested fix

In the `li` handler, detect block-level children (`pre`, `ol`, `ul`, `blockquote`, `table`) and:
1. Insert a `\n` before them (to break from inline text)
2. Indent their content to align with the list prefix width

Related: #114 (nested list indentation, different issue)

### PR DESCRIPTION
**Pull request by ManUtopiK**
## Summary

Fix two markdown rendering bugs in `renderMarkdown`:

- **#114 — Nested list indentation:** Nested lists now use the correct indentation width matching the parent list marker (`3` spaces for `1. `, `2` spaces for `- `), conforming to [CommonMark 0.31.2 §5.2](https://spec.commonmark.org/0.31.2/#list-items) — continuation blocks must be indented to the column of the first non-blank character after the list marker (Examples [254](https://spec.commonmark.org/0.31.2/#example-254), [109](https://spec.commonmark.org/0.31.2/#example-109)).
- **#118 — Block content in list items:** Block-level elements (`pre`, `blockquote`, `table`) inside list items are now rendered on their own line with proper indentation, instead of being concatenated to the preceding inline text. Per [CommonMark 0.31.2 §5.2 Example 263](https://spec.commonmark.org/0.31.2/#example-263), fenced code blocks in list items must be indented to align with list content.

## Changes

| File | Change |
| ---- | ------ |
| `indent.ts`            | Add `width` parameter for character-count indentation |
| `ul.ts`                | Set `listIndent: 2` in context (matches `- ` prefix)  |
| `ol.ts`                | Set `listIndent: 3` in context (matches `1. ` prefix) |
| `li.ts`                | Detect block-level children, render them on a new line with prefix-aligned indentation |
| 3 SPEC files           | Update expected markdown output to match corrected behavior |

## Test plan

- [x] All existing tests pass (746 passed, 2 pre-existing shiki timeouts unrelated)
- [x] Roundtrip stable: `parse → renderMarkdown → parse → renderMarkdown` produces identical output
- [x] Verified against CommonMark spec §5.2: indentation matches the column of first content after list marker

Closes #114, closes #118

### DIFF
--- a/packages/comark/src/internal/stringify/handlers/li.ts
+++ b/packages/comark/src/internal/stringify/handlers/li.ts
@@ -1,5 +1,10 @@
 import type { State } from 'comark/render'
 import type { ComarkElement, ComarkNode } from 'comark'
+import { indent } from '../indent.ts'
+
+// Block elements that need explicit indentation in list items.
+// Note: ol/ul are handled by their own handlers which manage indentation via listIndent context.
+const blockElements = new Set(['pre', 'blockquote', 'table'])
 
 export async function li(node: ComarkElement, state: State) {
   const children = node.slice(2) as ComarkNode[]
@@ -18,9 +23,21 @@ export async function li(node: ComarkElement, state: State) {
     prefix += input[1].checked || input[1][':checked'] ? '[x] ' : '[ ] '
   }
 
+  const prefixWidth = prefix.length
   let result = ''
+
   for (const child of children) {
-    result += await state.one(child, state, node)
+    const rendered = await state.one(child, state, node)
+
+    if (Array.isArray(child) && blockElements.has(child[0] as string)) {
+      // Block-level child: put on its own line and indent to align with list prefix
+      const trimmed = rendered.replace(/\n\n$/, '\n').replace(/\n$/, '')
+      const indented = indent(trimmed, { width: prefixWidth })
+      result = result.trimEnd() + '\n' + indented + '\n'
+    }
+    else {
+      result += rendered
+    }
   }
   result = result.trim()
 
--- a/packages/comark/src/internal/stringify/handlers/li.ts
+++ b/packages/comark/src/internal/stringify/handlers/li.ts
@@ -32,7 +32,7 @@ export async function li(node: ComarkElement, state: State) {
     if (Array.isArray(child) && blockElements.has(child[0] as string)) {
       // Block-level child: put on its own line and indent to align with list prefix
       const indented = indent(rendered, { width: prefixWidth })
-      result = result.trimEnd() + '\n' + indented + '\n'
+      result = result.trimEnd() + '\n' + indented.trimEnd() + '\n'
     }
     else {
       result += rendered
--- a/packages/comark/src/internal/stringify/handlers/li.ts
+++ b/packages/comark/src/internal/stringify/handlers/li.ts
@@ -31,8 +31,7 @@ export async function li(node: ComarkElement, state: State) {
 
     if (Array.isArray(child) && blockElements.has(child[0] as string)) {
       // Block-level child: put on its own line and indent to align with list prefix
-      const trimmed = rendered.replace(/\n\n$/, '\n').replace(/\n$/, '')
-      const indented = indent(trimmed, { width: prefixWidth })
+      const indented = indent(rendered, { width: prefixWidth })
       result = result.trimEnd() + '\n' + indented + '\n'
     }
     else {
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: packages/comark/src/internal/stringify/handlers/li.ts [hunk 1]
```diff
@@ -18,9 +23,21 @@ export async function li(node: ComarkElement, state: State) {
...
+  const prefixWidth = prefix.length
   let result = ''
+
   for (const child of children) {
-    result += await state.one(child, state, node)
+    const rendered = await state.one(child, state, node)
+
+    if (Array.isArray(child) && blockElements.has(child[0] as string)) {
+      // Block-level child: put on its own line and indent to align with list prefix
+      const trimmed = rendered.replace(/\n\n$/, '\n').replace(/\n$/, '')
```
farnabaz: Are you sure that these replaces are needed?
I just checked and even with removing these replaces tests passed.

I didn't get the real reason behind these replaces, maybe these are useful when li have multiple chilred.

Could you add a test to check against these replaces? 🙏

farnabaz:
Thanks for the PR @ManUtopiK 
I just add a comment on the li handler, other than that it looks great 🙂

packages/comark/src/internal/stringify/handlers/li.ts
ManUtopiK: You're right!
I tested with and without the .replace() and the output is the same in all cases. I started this current fix for some experiments with markdownlint. I didn't remember why I added this lines... so I removed them.

To be sure, I also added 7 SPEC tests to cover all cases (blockquote, table, code blocks, nested ol/ul up to 3 levels) here : https://github.com/ManUtopiK/comark/commit/f4c23309c2ae4f69863fd15ca26ebd6c1386198c

Should I merge these tests ? Is it too much ? Or in another PR ?

farnabaz: It's never too much to have more tests 🙂 
Feel free to push it in this PR 👍

ManUtopiK: C'est fait !

farnabaz (approved):
Thanks

PR #144, opened 2026-04-21
feat(punctuation): enhance punctuation plugin with normalization and locale-aware quotes

arashsheyda:
# Description

as discussed with @farnabaz `markdown-it` has a builtin support for punctuation (typography), this PR adds the missing functions and also you cna run the benchmark to see which is fast (our plugin is 2x-ish faster)


packages/comark/benchmark-punctuation.ts
```diff
@@ -0,0 +1,131 @@
+import { bench, run, group } from 'mitata'
```
arashsheyda: I should remove this file before merging, right?

atinux: I guess so, or maybe we could have a benchmark directory within the tests (or to the root) so we can keep track of the comark perf against others, so we can catch perf regression as the project grows

farnabaz (approved):

PR #148, opened 2026-04-21
fix(img): prevent undefined alt text in image markdown output

### LINKED ISSUE #146: Comark writes missing image alt texts as `undefined` strings
We just built a tiny internal syntax normalizer for nuxt-studio by using comarks parse & render APIs.

I noticed that comark is writing `undefined` into the alt texts of images that have nothing in the brackets in the source file.

```
![](/my/cool/path)
```
is turned into
```
![undefined](/my/cool/path)
```
after parsing and rendering back to markdown / mdc.

This screenshot is taken from our code after applying the normalizer we wrote:
Image

hendrikheil:
## Summary

Fixes #146.

When an image has no alt text (``), the `alt` attribute is absent from the AST node (the parser only sets it when `token.content` is truthy). The markdown renderer then used `alt` directly in a template literal, causing JavaScript to stringify `undefined` as the literal string `"undefined"`.

- Default `alt` to `''` in the destructuring in `packages/comark/src/internal/stringify/handlers/img.ts`
- Add SPEC test case `image-no-alt.md` to verify the round-trip produces `` not `undefined`

### DIFF
--- a/packages/comark/SPEC/COMARK/image-no-alt.md
+++ b/packages/comark/SPEC/COMARK/image-no-alt.md
@@ -1,0 +1,38 @@
+## Input
+
+```md
+![](/my/cool/path)
+```
+
+## AST
+
+```json
+{
+  "frontmatter": {},
+  "meta": {},
+  "nodes": [
+    [
+      "p",
+      {},
+      [
+        "img",
+        {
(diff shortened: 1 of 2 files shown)


farnabaz (approved):
Thanks ❤️

PR #149, opened 2026-04-21
feat(comark): add breaks plugin

maximepvrt:
It is a direct port of [`remark-breaks`](https://github.com/remarkjs/remark-breaks), designed to turn soft line breaks `\n` into `<br>` components. This allows users to have line breaks in the rendered output without requiring add `:br` or backslash at the end of the line.


farnabaz (approved):
Thanks ❤️

PR #151, opened 2026-04-22
chore(deps): clean up dependencies

arashsheyda:
$ Description

small pr to clean up the dependencies by removing the unused ones:

- removed `@nuxt/kit` from devDependencies, not imported in any source or test file
- removed `@shikijs/primitive` from devDependencies, was previously used but refactored to import from `shiki` directly
- removed `shiki` from dependencies, it was never imported in any source file. users who need highlighting already get it via comark's optional peer dependency


farnabaz (approved):

farnabaz (approved):
Thanks ❤️

PR #159, opened 2026-04-23
chore: use oxlint and oxfmt

arashsheyda:
# Description

this PR replaces `eslint` with oxlint and oxfmt for better performances: ~4.8x faster overall 🤯


farnabaz:
Great PR 👍 

It would e great if we can keep this rule too.
```
  .append({
    files: ['packages/**/src/**'],
    rules: {
      'import/extensions': [
        'error',
        'ignorePackages',
      ],
    },
  })
```

Is there any option in oxlint to enforce having extension for relative imports?
It is important to have extension in relative imports, otherwise the package will be broken.

arashsheyda:
Ah yes, https://oxc.rs/docs/guide/usage/linter/rules/import/extensions.html#import-extensions

farnabaz (approved):
Nice work 👍 
Thanks ❤️

PR #162, opened 2026-04-24
perf(plugins): optimize highlight code processing

arashsheyda:
### ❓ Type of change

### 📚 Description

while working on the benchmark file, I noticed that our highlight plugin is significantly slower than others. this PR optimizes the plugin to eliminate overhead in the hot path brining from `~1.65 ms` down to `~175 µs` for small inputs

| Before | After |
|--------|--------|
| Screenshot 2026-04-24 at 12 18 46 PM | Screenshot 2026-04-24 at 12 16 30 PM | 

#### What changed

1. `getHighlighter()`
If the singleton highlighter is already there and we’re not passing custom themes/languages (which is most cases), we just return it.
Before, it would still go through registerDefaults() and await a bunch of dynamic imports + theme/language loads… even when everything was already loaded.

2. `hastToMinimarkNode()`: no more extra allocations
Got rid of the spread + .map() combo and replaced it with a pre-allocated array + simple for-loop.
Previously we were creating a mapped array and then spreading it into another array — so 2 allocations per node.

Also moved it out of highlightCode so it’s not recreated on every call.

3. `highlightCodeBlocks()`: no async overhead
codeToHast (from Shiki v4) is actually synchronous.
We were wrapping it in Promise.all(...map(async ...)), which just added unnecessary microtask overhead per block.

Now it’s just a plain loop.

4. `findCodeBlocks()`: zero-allocation traversal
The old recursion was creating a new path array and slicing children on every step.

Switched to a shared pathBuf with push/pop, and just iterating children starting from index 2.
No more extra arrays per recursion.

5. Highlight line lookup: using a Set
Swapped .includes() (O(n)) for a Set (O(1)).
Simple win since this runs per line.

6. Code node construction: pre-allocated array
Instead of spreading potentially hundreds of children (['code', attrs, ...children]),
we allocate once and fill it with a loop.

7. Class string building: simpler concat
Replaced the array + .filter().join() with direct string concatenation and a precomputed suffix.
Less work, same result.

8. Tree cloning — only copy what we touch
Got rid of JSON.parse(JSON.stringify(...)).

Now we only copy the path (spine) down to the modified `<pre>` nodes, and keep the rest of the tree shared.
Way less memory churn.

#### What didn’t change

- Public API: everything stays the same (highlightCode, highlightCodeBlocks, etc.)
- Output: exact same tree structure, tags, attrs, classes, ordering
- Line highlighting behaviuor: unchanged
- Theme/language loading: same defaults + lazy loading
- Error handling: still falls back to raw code if highlighting fails

### 📝 Checklist

### DIFF
--- a/benchmarks/plugin-highlight.ts
+++ b/benchmarks/plugin-highlight.ts
@@ -1,0 +1,145 @@
+import { bench, run, group, barplot } from 'mitata'
+import MarkdownIt from 'markdown-it'
+import MarkdownExit from 'markdown-exit'
+import pluginMdc from '@comark/markdown-it'
+import { createParse } from 'comark'
+import highlight, { getHighlighter } from '../packages/comark/src/plugins/highlight'
+import { codeToHast } from 'shiki/core'
+
+const short = `
+# Quick Start
+
+Install with:
+
+\`\`\`bash
+npm install comark
+\`\`\`
+
+Then use it:
+
+\`\`\`javascript
+import { parse } from 'comark'
+const tree = await parse('# Hello')
+\`\`\`
+`
+
+const medium = `
+# API Reference
+
+## Parse
+
+\`\`\`typescript
+import { parse } from 'comark'
+
+interface ParseOptions {
+  autoClose?: boolean
+  streaming?: boolean
+  plugins?: ComarkPlugin[]
+}
+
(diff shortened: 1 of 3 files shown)


farnabaz (approved):
Great work 👌 
LGTM 👍

PR #169, opened 2026-04-28
fix(stringify): handle multiple paragraphs in a list item

### LINKED ISSUE #167: bug: Single paragraph in list item unwrapped even if autoUnwrap is false
### Environment

--

### Comark Version

v0.3.1

### Reproduction

https://github.com/annaghi/comark-repro-auto-unwrap-ignored

### Description

Paragraph unwrapping in list items occurs in two places

1. `auto-unwrap.ts`: controlled by a feature flag
2. `token-processor.ts`: runs unconditionally

`token-processor.ts` always unwraps a single-child paragraph in a list item, regardless of whether `autoUnwrap` is enabled. Setting the flag to `false` has no effect on this case.

### Additional context

--

### Logs

```shell-script
[
  [
    'h1',
    { id: 'unwrapping-ignores-the-auto-unwrap-flag' },
    'Unwrapping ignores the auto-unwrap flag'
  ],
  [ 'ul', {}, [ 'li', {}, 'para' ] ]
]
{ options: { autoUnwrap: false } }
```

### LINKED ISSUE #168: bug: Two-paragraph list items are rendered as inline content
### Environment

--

### Comark Version

v0.3.1

### Reproduction

https://github.com/annaghi/comark-repro-token-processor-list-item-multi-paragraph

### Description

When a list item contains two paragraphs, `token-processor.ts` unwraps both, stripping the block structure. The result is inline elements where block elements are expected, producing incorrect HTML output.

### Additional context

--

### Logs

```shell-script
[
  [
    'h1',
    { id: 'two-paragraph-list-items-are-rendered-as-inline-content' },
    'Two-paragraph list items are rendered as inline content'
  ],
  [ 'ul', {}, [ 'li', {}, 'para1', 'para2' ] ]
]
```

farnabaz:
### 🔗 Linked issue

resolves #168
resolves #167

### ❓ Type of change

### 📚 Description

- If `<li>` contains only one paragraph -> the auto unwrap feature will remove that paragraph an unwrap its children
- If the `<li>` contains more than one child parser will preserve all tags

### 📝 Checklist


annaghi:
This was super fast, thank you! :sparkles: Multi-paragraph list-item renders correctly.

While reviewing this PR, I realized that I was wrong about the single-paragraph list item requiring `p` wrapper, IF Comark adheres CommonMark. `auto-unwrap` should not affect single-paragraph list items, since they do not have a wrapper - according to CommonMark.

On the other hand, this PR makes wrapping and unwrapping consistent: either always wrap into `p` or always unwrap, rather than mixing the two approaches as Commonmark does. For example:

### CommonMark

- single paragraph list item: `li { text }`
- single paragraph blockquote: `blockquote { p { text } }`

### With this PR

`autoUnwrap : true`

- single paragraph list item: `li { text }`
- single paragraph blockquote: `blockquote { text }`

`autoUnwrap : false`

- single paragraph list item: `li { p { text } }`
- single paragraph blockquote: `blockquote { p { text } }`

Nice, consistent approach! :fire:

PR #170, opened 2026-04-28
perf(plugins): optimize highlight to use codeToTokens fast path to skip hast intermediary

arashsheyda:
### ❓ Type of change

### 📚 Description

this PR optimizes the highlight plugin further by using shiki's `codeToTokens` API directly instead of `codeToHast` when no transformers are configured. this skips the entire hast tree construction and the subsequent `hast => ComarkNode` conversion, building ComarkNodes directly from the flat token array in a single pass

| Before | After |
|--------|--------|
| Screenshot 2026-04-28 at 11 43 52 AM | Screenshot 2026-04-28 at 11 45 49 AM | 

### 📝 Checklist


farnabaz (approved):
Comark used to use code to tokens at first but in order to support transformers we moved to code to hast.
Switching between these two based on transformers existence is smart move 👍

PR #183, opened 2026-05-06
fix(svelte): add missing components-manifest prop

### LINKED ISSUE #177: `componentsManifest` is missing in @comark/svelte
### Environment

- @comark/svelte: 0.3.1
- Framework: Svelte

### Comark Version

0.3.1

### Reproduction

https://comark.dev/rendering/svelte#code-comarkrenderer-code-componentsmanifest

### Description

Hi there. First of all, I want to thank you for such a well-designed package. It's really cool!

As for the issue, the documentation states that you can lazily load components on demand using `componentsManifest`. But based on the [source code](https://github.com/comarkdown/comark/blob/main/packages/comark-svelte/src/components/ComarkNode.svelte), this prop doesn't do anything at all.

atinux:
### 🔗 Linked issue

resolves #177

### ❓ Type of change

### 📚 Description

We did not have the `components-manifest` prop on Svelte's comark components.

This PR aims to add them, but my lack of knowledge on Svelte makes me unconfortable right now, I would love a review from @elliott-with-the-longest-name-on-github first 🙏 

Maybe @Serator as well?

### 📝 Checklist


elliott-with-the-longest-name-on-github (commented):
I'm out of town until Monday but I do know that at least as is current, the `import.meta.glob` example does not lazily load components (unless something has changed on the Vite end). `import.meta.glob` basically performs a buildtime lookup of all available files and inlines them as imports into the file. 

I'll take a better look at this on Monday!

atinux:
@elliott-with-the-longest-name-on-github actually the `eager: true` strategy make sure to not lazy-load the chunks, see https://vite.dev/guide/features#glob-import

packages/comark-svelte/src/components/ComarkNode.svelte
```diff
@@ -14,10 +14,53 @@ naturally appears inline after the deepest trailing text node.
...
+  function resolveComponent(tag: string, components: Record<string, any>, componentsManifest?: ComarkComponentManifest): any {
+    const pascal = pascalCase(tag)
+    const Component
+      = components[`Prose${pascal}`]
+        || components[pascal]
+        || components[tag]
+
+    if (Component || !componentsManifest) {
+      return Component || null
+    }
```
elliott-with-the-longest-name-on-github: nit but this is kind of difficult to understand. with a little more code it becomes much clearer:
```suggestion
    if (Component) {
      return Component
    }
    if (!componentsManifest) {
      return null;
    }
```

packages/comark-svelte/src/components/ComarkNode.svelte
```diff
@@ -105,35 +148,37 @@ naturally appears inline after the deepest trailing text node.
...
-        {componentsManifest}
-        caretClass={i === children.length - 1 ? caretClass : null}
-        renderData={childrenRenderData}
-      />
-    {/each}
+    {@render renderChildren()}
   </Component>
+{:else if componentPromise}
+  {#await componentPromise then AsyncComponent}
+    <AsyncComponent {...mappedProps}>
```
elliott-with-the-longest-name-on-github: FWIW this will render empty content during SSR. I don't think we can do better than that without experimental  async svelte but figured I'd call it out.

elliott-with-the-longest-name-on-github: Oh. Actually this is rather annoying. Since this is shared between non-experimental and `ComarkAsync`, it won't render anything in `ComarkAsync` either. Hmm. I think probably the correct way to do this would be to make a `Resolve` and `ResolveAsync` component that `ComarkAsync` and `Comark` can pass in, respectively. `ResolveAsync` would basically look like:

```svelte
<script>
  let { promise, props, children } = $props();
  const Component = $derived(await promise);
</script>

<Component {...props}>
  {@render children?.()}
</Component>
```

It's such a tiny component in both cases that it wouldn't hurt to make `Resolve` the default so that `ComarkNode` doesn't actually have to have its consumer pass in the value. `ComarkAsync` can pay the price of having the useless `Resolve` when it overrides it with `ResolveAsync`.

elliott-with-the-longest-name-on-github (commented):
LGTM, just one nit and one slightly more substantial change to support experimental async

Serator:
Hey @atinux. I don't know how it happened, but I missed the notification. 🤷 @elliott-with-the-longest-name-on-github  has already mentioned `import.meta.glob` [earlier](https://github.com/comarkdown/comark/pull/183#pullrequestreview-4257470937). `import.meta.glob(‘./components/*.svelte’, {eager: true})` will cause all components under the specified path to be loaded; in other words, they will be included in the final bundle even if they are not used. I tried your PR in SSG mode, and in my case, that's exactly how it is.

A few words about my case:

I have some Markdown pages, some of which contain custom widgets. I'd like to be able to store these widgets directly in the pages' directories and load them during the build (SSG). Something like:

```Markdown
...

![Some widget](./SomeWidget/index.svelte)

...
```

(The path to the widget may or may not be unique)

With `import.meta.glob`, I would have to specify the path in advance (In fact, [you cannot use variables in it](https://vite.dev/guide/features#glob-import-caveats), as is done in the example with the [current documentation](https://comark.dev/rendering/svelte#code-comarkrenderer-code-componentsmanifest)), which completely defeats the purpose of loading the component on demand (in my case).

atinux:
Alright, I implemented both `Resolve` and `ResolveAsync` components + added a SvelteKit example

atinux:
@Serator you don't need to use `import.meta.glob` but directly from the page where you need to lazy-load the components that will be rendered in your Markdown to create your componentsMap:


```svelte
<script lang="ts">
  import { ComarkAsync } from '@comark/svelte/async'
  import type { PageData } from './$types'

  let { data }: { data: PageData } = $props()

  const componentMap: Record<string, () => Promise<any>> = {
    'alert': () => import('$lib/components/Alert.svelte'),
    'card': () => import('$lib/components/Card.svelte'),
  }

  const componentsManifest = (name: string) => componentMap[name]?.()
</script>

<article>
  <ComarkAsync markdown={data.markdown} {componentsManifest} />
</article>
```

See https://comark.dev/examples/frameworks/sveltekit

farnabaz (approved):
👍

larbish (approved):

PR #184, opened 2026-05-07
fix(syntax): thematic breaks in component slots

larbish:
Fix `---` inside block components being parsed as YAML props.

Only treat `---` as props when it directly follows the component opener, otherwise it's a thematic break.


atinux:
Could you add a test @larbish ?

larbish:
@atinux SPEC markdown files are tests!

farnabaz (approved):
Thanks

PR #187, opened 2026-05-11
fix(plugins): headings plugin has remove false by default

atinux:
### ❓ Type of change

### 📚 Description

- Fix docs and JSDoc that incorrectly stated `remove` defaults to `true`, the actual default is `false` since the `headings: do not remove extracted node` change
- Add `headings.test.ts` covering default extraction, `remove: true`, missing title, and custom tags

### 📝 Checklist


farnabaz (approved):
Thanks

PR #198, opened 2026-05-18
feat(parse): infer typed meta/frontmatter from plugins

### LINKED ISSUE #196: Type narrowing for the types returned by the `parse` function (for `frontmatter` / `meta`)
### Description

Currently, `frontmatter` / `meta` are always returned as `Record<string, any>`. This type can be narrowed using type casting or a type guards, but it requires duplicating code and is prone to errors.

It would be great to be able to generate typed output. If it’s not possible to type the entire output, then at least the part that is processed by Comark plugins.

### Motivation

Understanding types will help write more robust code.

For example, when we use `result.meta.toc` from the [ToC plugin](https://comark.dev/plugins/built-in/toc), we get `any`. 

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Additional context

_No response_

farnabaz:
### 🔗 Linked issue
resolves #196

### ❓ Type of change

### 📚 Description
Plugins can extend the type

```ts
// headings plugin
export default defineComarkPlugin<HeadingsOptions, { title?: string; description?: string }>((options = {}) => {
}
```
And you can see result type in the image
Screenshot 2026-05-18 at 16 20 16
Screenshot 2026-05-18 at 16 22 41

### 📝 Checklist

### DIFF
--- a/packages/comark/src/plugins/toc.ts
+++ b/packages/comark/src/plugins/toc.ts
@@ -136,7 +136,7 @@ export function generateFlatToc(body: ComarkTree, options: Toc): Toc {
   }
 }
 
-export default defineComarkPlugin((options: Partial<Toc> = {}) => {
+export default defineComarkPlugin<Partial<Toc>, { toc: Toc }>((options: Partial<Toc> = {}) => {
   const { title = '', depth = 2, searchDepth = 2, links = [] } = options
   return {
     name: 'toc',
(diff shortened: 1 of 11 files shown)


packages/comark/src/plugins/toc.ts
Serator, line 147: @farnabaz Is it possible for the plugin itself to respect the types as well? In this case, we won't get any type errors despite the types specified above.

```
state.tree.meta.toc = 123
```

farnabaz: Great one 👍  Just pushed a commit to support this :)

atinux:
Looks good to me, can we please document it as well in https://comark.dev/plugins/custom/plugin-api ?

atinux (approved):

PR #208, opened 2026-05-21
fix: parse indented raw html children in html blocks

atinux:
## Why
Indented raw HTML children inside an open HTML block could be tokenized as markdown code blocks. 

```md
<p align="center">
  <a href="https://cdn.jsdelivr.net/gh/antfu/static/sponsors.svg">
    <img src="https://cdn.jsdelivr.net/gh/antfu/static/sponsors.svg" alt="Sponsors"/>
  </a>
</p>
```

was creating:

```ts
[
  [
    "p",
    { "$": { "html": 1, "block": 1 }, "align": "center" },
    [
      "a",
      {
        "$": { "html": 1, "block": 1 },
        "href": "https://cdn.jsdelivr.net/gh/antfu/static/sponsors.svg"
      }
    ],
    [
      "pre",
      {},
      [
        "code",
        {},
        "<img src=\"https://cdn.jsdelivr.net/gh/antfu/static/sponsors.svg\" alt=\"Sponsors\"/>"
      ]
    ]
  ],
  ["component", {}]
]
```

## What
- Process multiline `html_block` tokens through one HTML-aware path.
- Reinterpret HTML-looking `code_block` tokens as raw HTML only while inside an HTML block.
- Keep normal markdown parsing inside HTML blocks, including emphasis, inline code, components, and real indented code blocks.
- Add regression coverage for nested indented raw HTML, markdown inside HTML, comments, stray close tokens, and code-block boundaries.

## Tests
- `pnpm --dir packages/comark exec vitest run test/html-block.test.ts`
- `pnpm --dir packages/comark test`
- `pnpm exec oxfmt --check packages/comark/test/html-block.test.ts`

Not run in this final pass: `pnpm run typecheck` due existing unrelated errors in `packages/comark-vue/test/vite.test.ts`.


atinux:
The more I think, the less I believe we should be able to mix markdown within html @farnabaz, this would probably simplify a lot the logic & performance

farnabaz (approved):

PR #212, opened 2026-05-22
fix: stricter boolean prop guessing

atinux:
This should fix the error on landing page when streaming and icon prop is wrongly considered as boolean for the `::callout`

The `'true'` is added before to we can have a stricter check here


larbish (approved):
New tests look legit ✅

farnabaz (approved):

PR #213, opened 2026-05-22
fix(syntax): nested codeblock inside slots

larbish:
Closing `::` inside a fenced code block in a **slotted** parent component was ending the parent component prematurely, causing the code block content to be re-parsed as component. 

`comark_block_slots` now mirrors the fence tracking already present in `comark_block` (`component-nested-codeblock.md` SPEC has been added to test the already working case without slot).


farnabaz (approved):
Nice Catch 👍

PR #214, opened 2026-05-22
feat(block-attributes): attach trailing `{attr}` to nearest block element

### LINKED ISSUE #222: Feature: new syntax for heading with custom attributes
### Description

https://github.com/nuxt-content/remark-mdc/issues/122

### Motivation

x

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Additional context

_No response_

farnabaz:
resolves #222

## Summary

Two related improvements to block-level attribute handling.

### 1. Attach trailing `{attr}` to the nearest block element

A trailing `{attr="value"}` on the last line of a block now attaches to that block. Works for paragraphs, headings (h1–h6), bullet/ordered/task list items, and blockquote paragraphs:

```md
A paragraph {attr="value"}

- a list item {attr="value"}
- another list item {attr2="value2"}

# h1 {attr="value"}
###### h6 {attr="value"}

> Blockquote {attr="value"}

> Para 1 {attr="value"}
>
> Para 2 {attr2="value2"}

1. List item {attr="value"}

- [ ] Task list item {attr="value"}
- [x] Task list item {attr2="value2"}
```

Implementation:

- The inline-props lift in `syntax.ts` is loosened to trigger when props follow a text node even if other inline siblings (e.g. the task-list checkbox) precede it, while still skipping markdown-it's empty filler text token between an emphasis close and `{...}` so `**bold**{attr}` keeps attaching to `<strong>` instead of the `<p>`.
- `heading_open` merges `token.attrs` with the auto-generated id.
- Auto-unwrap lifts the unwrapped paragraph's attrs onto the parent (parent attrs win), so `> Blockquote {attr}` keeps the attr after the paragraph is removed.
- Stringifiers (p / heading / blockquote / li) emit the trailing `{attr=…}` suffix; auto-generated heading id, blockquote `as`, and the `task-list-item` class are filtered out so they don't echo back as user attrs.

### 2. Fold `::ul` / `::ol` / `::table` / `::blockquote` into their single same-tag child

The native markdown syntax has no slot for list-level, table-level, or multi-paragraph-blockquote-level attrs. To cover those, the matching `::tag` block component is now a first-class way to attach attrs to the underlying element. When `::ul` / `::ol` / `::table` / `::blockquote` wraps exactly one same-tagged child, the parser folds them into a single element — the wrapper's attrs land on the actual list/table/blockquote element instead of producing a nested duplicate.

```md
::ul{attr="value"}
- item 1
- item 2
::
```

parses to `["ul", {attr: "value"}, [li], [li]]` (not `["ul", {attr}, ["ul", {}, ...]]`), and renders to `<ul attr="value">...</ul>`. Same shape for `::ol`, `::table`, and `::blockquote`. Outer attrs win over inner if both are present.

For the markdown roundtrip, ul / ol / table emit the `::tag{attrs}\n…\n::` form when they carry user attrs. Blockquote uses the wrapper form when it has user attrs *and* multi-block children; single-content blockquotes keep the natural `> text {attr}` form. The old markdown-it-level `::ul` dedupe hook was removed since the AST-level fold supersedes it.

### DIFF
--- a/docs/app/composables/useNavigation.ts
+++ b/docs/app/composables/useNavigation.ts
@@ -36,6 +36,12 @@ export function useMainNavigation() {
         active: route.path.startsWith(`/play/${example.value}`),
       })),
     },
+    {
+      label: 'Compare',
+      to: '/compare',
+      icon: 'i-lucide-columns-2',
+      active: route.path.startsWith('/compare'),
+    },
   ])
 }
 
--- a/docs/app/composables/useNavigation.ts
+++ b/docs/app/composables/useNavigation.ts
@@ -36,12 +36,6 @@ export function useMainNavigation() {
         active: route.path.startsWith(`/play/${example.value}`),
       })),
     },
-    {
-      label: 'Compare',
-      to: '/compare',
-      icon: 'i-lucide-columns-2',
-      active: route.path.startsWith('/compare'),
-    },
   ])
 }
 
--- a/docs/app/pages/compare.vue
+++ b/docs/app/pages/compare.vue
@@ -1,0 +1,105 @@
+<script setup lang="ts">
+import { useLocalStorage } from '@vueuse/core'
+import { allFeaturesMarkdown } from '~/constants'
+
+definePageMeta({ footer: false })
+
+useSeoMeta({
+  title: 'Compare - Comark',
+  description: 'Compare multiple Comark documents side by side.',
+})
+
+interface Col {
+  id: string
+  content: string
+}
+
+const defaultCols: Col[] = [
+  { id: '1', content: allFeaturesMarkdown },
+  { id: '2', content: allFeaturesMarkdown },
+]
+
+const cols = useLocalStorage<Col[]>('comark:compare-cols', defaultCols, {
+  serializer: {
+    read: (v) => {
+      try {
+        const parsed = JSON.parse(v)
+        return Array.isArray(parsed) && parsed.length >= 2 ? parsed : defaultCols
+      } catch {
+        return defaultCols
+      }
+    },
+    write: (v) => JSON.stringify(v),
+  },
+})
+
+// nextId must be higher than any existing id to avoid collisions after reload
+let nextId = cols.value.reduce((max, c) => Math.max(max, Number(c.id) || 0), 0) + 1
+
+function addCol() {
+  cols.value = [...cols.value, { id: String(nextId++), content: '' }]
+}
+
+function removeCol(id: string) {
+  if (cols.value.length <= 2) return
+  cols.value = cols.value.filter((c) => c.id !== id)
+}
+</script>
+
+<template>
+  <ClientOnly>
+    <div class="flex h-[calc(100vh-64px)] overflow-hidden">
+      <!-- Editor columns -->
+      <div
+        v-for="(col, i) in cols"
+        :key="col.id"
+        class="flex min-w-0 flex-1 flex-col border-l border-default first:border-l-0"
+      >
+        <div class="flex h-9 shrink-0 items-center gap-2 border-b border-default bg-elevated px-3">
+          <span class="text-xs text-muted">Editor {{ i + 1 }}</span>
+          <UButton
+            v-if="cols.length > 2"
+            icon="i-lucide-x"
+            variant="ghost"
+            color="neutral"
+            size="xs"
+            class="ml-auto -mr-1"
+            @click="removeCol(col.id)"
+          />
+        </div>
+        <div class="min-h-0 flex-1">
+          <Editor v-model="col.content" />
+        </div>
+      </div>
+
+      <!-- Add column -->
+      <div class="flex w-10 shrink-0 items-center justify-center border-l border-default bg-elevated">
+        <UTooltip text="Add editor">
+          <UButton
+            icon="i-lucide-plus"
+            variant="ghost"
+            color="neutral"
+            size="sm"
+            @click="addCol"
+          />
+        </UTooltip>
+      </div>
+    </div>
+
+    <template #fallback>
+      <div class="flex h-[calc(100vh-64px)] overflow-hidden">
+        <div
+          v-for="i in 2"
+          :key="i"
+          class="flex flex-1 flex-col border-l border-default first:border-l-0"
+        >
+          <div class="h-9 border-b border-default bg-elevated" />
+          <div class="flex-1 p-4">
+            <USkeleton class="h-full w-full" />
+          </div>
+        </div>
+        <div class="w-10 shrink-0 border-l border-default bg-elevated" />
+      </div>
+    </template>
+  </ClientOnly>
+</template>
--- a/packages/comark/SPEC/COMARK/attributes/blockquote-multiple-paragraphs.md
+++ b/packages/comark/SPEC/COMARK/attributes/blockquote-multiple-paragraphs.md
@@ -1,0 +1,53 @@
+## Input
+
+```md
+> Blockquote paragraph 1 {attr="value"}
+>
+> Blockquote paragraph 2 {attr2="value2"}
+```
+
+## AST
+
+```json
+{
+  "frontmatter": {},
+  "meta": {},
+  "nodes": [
+    [
+      "blockquote",
+      {},
+      [
+        "p",
+        {
+          "attr": "value"
+        },
+        "Blockquote paragraph 1"
+      ],
+      [
+        "p",
+        {
+          "attr2": "value2"
+        },
+        "Blockquote paragraph 2"
+      ]
+    ]
+  ]
+}
+```
+
+## HTML
+
+```html
+<blockquote>
+  <p attr="value">Blockquote paragraph 1</p>
+  <p attr2="value2">Blockquote paragraph 2</p>
+</blockquote>
+```
+
+## Markdown
+
+```md
+> Blockquote paragraph 1 {attr="value"}
+>
+> Blockquote paragraph 2 {attr2="value2"}
+```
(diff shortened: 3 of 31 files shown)


annaghi:
Hello, this is a great feature to have!

I have seen block attributes being implemented as leading attributes, and if you have time can you explain why Comark needs them to be trailing ones? (perhaps because of the streaming?)

This is the same markdown as yours, with block attributes for `ul` and `blockquote` (`auto-unwrap` is not applied). I see that this approach has twice as long markdown file, but it does not need the special **Fold** case:

```md
{attr="value"}
A paragraph

- {attr="value"}
  a list item
- {attr2="value2"}
  {attr3="value3"}
  another list item

{attr="value"}
# h1

{attr="value"}
###### h6

{attr="value"}
> Blockquote

> {attr="value"}
> Para 1
>
> {attr2="value2"}
> Para 2

1. {attr="value"}
  List item

{attr="value"}
- {attr1="value1"}
  [ ] Task list item
- {attr2="value2"}
  [x] Task list item
```

```html
<p attr="value">A paragraph</p>

<ul>
  <li attr="value">
    <p>a list item</p>
  </li>
  <li attr2="value2">
    <p attr3=""value3"">another list item</p>
  </li>
</ul>

<h1 attr="value">h1</h1>
<h6 attr="value">h6</h6>

<blockquote attr="value">
  <p>Blockquote</p>
</blockquote>

<blockquote>
  <p attr="value">Para 1</p>
  <p attr2="value2">Para 2</p>
</blockquote>

<ol>
  <li attr="value">
    <p>List item</p>
  </li>
</ol>

<ul attr="value">
  <li attr1="value1">
    <p>[ ] Task list item</p>
  </li>
  <li attr2="value2">
    <p>[x] Task list item</p>
  </li>
</ul>
```

farnabaz:
> Hello, this is a great feature to have!
> 
> I have seen block attributes being implemented as leading attributes, and if you have time can you explain why Comark needs them to be trailing ones? (perhaps because of the streaming?)
> 
> This is the same markdown as yours, with block attributes for `ul` and `blockquote` (`auto-unwrap` is not applied). I see that this approach has twice as long markdown file, but it does not need the special **Fold** case:


One of the keys things that we have in mind while developing new feature is to keep Markdown soul (easy-to-read, easy-to-write). Although prefix lines seems  a good unified extension for markdown itself but it affect the readability of markdown.

You can see in the following images, the third markdown is basically the same markdown we see day to day and contents didn't shifted.

Screenshot 2026-05-25 at 16 17 14

Another important note for Comark is to render well in Common Mark renderers. For example Github renderer which we use day to day is an important metric for this. See image bellow

Screenshot 2026-05-25 at 16 17 02

In these two image, the third column (This proposal) preserve positioning and your eye will not look around to find the start point of actual content.


As for wrapper components,  the very same reason applies but since Comark provides a component syntax, we decided to  leverage that feature and let auto-unwrap tackle with the duplication. 

Complexity Note:
Markdown renderer are mainly have linear approach to parse content. Having attributes as trailing has smaller overhead to parser since the attribute's owner node is already created. While in leading version, parser should keep attributes in memory and attach to upcoming node. Although this is not a big deal performance wise but it will need another loop through the whole Nodes Tree.

annaghi:
Thank you for the detailed explanation, I really appreciate it.

Now the benefits are clear: keeping the markdown shape familiar for authors, and keeping the output familiar / readable using renderers which do not support attributes.

These examples also helped me to narrow down my question: perhaps I need to distinguish the different node types: headings having trailing attributes feels natural, but for example currently we cannot add attributes to a list item which has multiple blocks inside (using CommonMark's lists).

About the implementation: currently I use a post-processor on the AST which looks for text of attributes and moves them to the next or parent block. (List items are problematic regardless.)

I know I have no say how this feature is implemented, that is fine :smile: but your perspective helps to shape the port I work on, and cool down my struggles of deviating from Comark or not. Thanks again! :sparkles:

farnabaz:
@annaghi If you have time to spare, I’d actually love to hear more about your POV on the feature.

Hearing different perspectives on these kinds of design and implementation tradeoffs is always valuable.

annaghi:
Absolutely. I am going to collect my thoughts, and post here.

annaghi:
Probably the simplest way is to show an example with just block attributes and Markdown (without directives or components):


```md
{.grid.cards}
- :material-clock-fast:{.twemoji.lg.middle} **Set up in 5 minutes**

  ---

  Install [`product`](#) with [`pm`](#) and get up and running in minutes

  [:octicons-arrow-right-24:{.twemoji}Getting started](#)

- :{.featured}

  :fontawesome-brands-markdown:{.twemoji.lg.middle} **It's just Markdown**

  ---

  Focus on your content and generate a responsive and searchable static site

  [:octicons-arrow-right-24:{.twemoji}Reference](#)

- :material-format-font:{.twemoji.lg.middle} **Made to measure**

  ---

  Change the colors, fonts, language, icons, logo and more with a few lines

  [:octicons-arrow-right-24:{.twemoji}Customization](#)

- :material-scale-balance:{.twemoji.lg.middle} **Open Source, MIT**

  ---

  Product is licensed under MIT and available on [GitHub](#)

  [:octicons-arrow-right-24:{.twemoji}License](#)
```

`svg` related things and `href`s were removed:

```html
<ul class="grid cards">
  <li>
    <p>
      <svg class="icon twemoji lg middle"></svg> <strong>Set up in 5 minutes</strong>
    </p>
    <hr>
    <p>Install <a href="#"><code>product</code></a> with <a href="#"><code>pm</code></a> and get up and running in minutes</p>
    <p>
      <a href="#"><svg class="icon twemoji"></svg>Getting started</a>
    </p>
  </li>
  
  <li class="featured">
    <p>
      <svg class="icon twemoji lg middle"></svg> <strong>It’s just Markdown</strong>
    </p>
    <hr>
    <p>Focus on your content and generate a responsive and searchable static site</p>
    <p>
      <a href="#"><svg class="icon twemoji"></svg>Reference</a>
    </p>
  </li>
  
  <li>
    <p>
      <svg class="icon twemoji lg middle"></svg> <strong>Made to measure</strong>
    </p>
    <hr>
    <p>Change the colors, fonts, language, icons, logo and more with a few lines</p>
    <p>
      <a href="#"><svg class="icon twemoji"></svg>Customization</a>
    </p>
  </li>
  
  <li>
    <p>
      <svg class="icon twemoji lg middle"></svg> <strong>Open Source, MIT</strong>
    </p>
    <hr>
    <p>Product is licensed under MIT and available on <a href="#">GitHub</a></p>
    <p>
      <a href="#"><svg class="icon twemoji"></svg>License</a>
    </p>
  </li>
</ul>
```

The rest is just raw CSS:

Screenshot from 2026-05-27 11-39-55

From semantic HTML point of view: it seems fine to have grids like cards or articles in `ul` instead of `div`, for example MDN does that: [cards](https://developer.mozilla.org/en-US/about) and [articles](https://developer.mozilla.org/en-US/).

There are two nuances on which my implementation relies on currently (it need to be revisited, since headings and maybe paragraphs too are going to have trailing attributes):

- need to use a colon on the list items `:{attrs}` (the colon was a choice of taste), because the AST is the same for
  ```md
  - {attrs}
    > qoute
  ```
  and
  ```md
  -
    {attrs}
    > qoute
  ```
  and I do not want to modify the upstream parser if I do not need to

- need to add a blank line before the nested block's attributes (probably this is unavoidable, because we need to start a new block, and blank line is the way to do it)

So currently this works:

```md
{.ul}
- :{.li}

  {.nested-ul}
  - :{.nested-li}
    [x] Nested task list
```

and this does not:

```md
{.ul}
- {.li}
  {.nested-ul}
  - {.nested-li}
    [x] Nested task list
```

Perhaps it is fine to have a special attribute syntax for list items `:{attrs}` because it makes the user intention explicit, I am not sure yet.

I am happy to hear more to refine the feature.

farnabaz:
Thanks for the explanation 🙂 
So from what I see, your goal is to design a syntax to create static pages with only markdown toolchain using props and attributes. Minimum requirement for div or other html tags. 

It make sense, but I still didn't get why did you choose leading attributes for block. It is only to prevent folding?

annaghi:
I realized that calling these attributes "leading" was misleading on my part, they are "separate line" attributes (as you call them).

Using separate-line-attribute syntax was not really a choice, I found this: https://github.com/jgm/djot

[docs](https://htmlpreview.github.io/?https://github.com/jgm/djot/blob/master/doc/syntax.html#block-attributes)

The list-item attributes was my invention, DJot does not cover that.

farnabaz:
Interesting, the author is also author of https://github.com/commonmark/commonmark-spec 👍 

Thanks for sharing, Now all parts fit together. list-item attributes are also nicely fit with djot spec and minor `:` difference is looks fine IMO 👍

annaghi:
However I have just noticed something in the Djot Readme.md, those leading spaces before the block attributes:

```md
                                                             {#introduction}
This is the introductory paragraph, with
an identifier `introduction`.

                                                             {.important color="blue" #heading}
## heading

The word *atelier*{weight="600"} is French.
```

A fun idea: The Markdown file in GitHub's preview looks still somewhat readable. If the VSCode extension could support the right alignment and colorization...

farnabaz:
> However I have just noticed something in the Djot Readme.md, those leading spaces before the block attributes:

To be honest, I'm afraid it also makes writing harder.  And it will render attribute blocks as code block in Github.

---


                                                             {#introduction}
This is the introductory paragraph, with
an identifier `introduction`.

                                                             {.important color="blue" #heading}
## heading

The word *atelier*{weight="600"} is French.

PR #228, opened 2026-06-03
chore: add sideEffects false to all packages, move katex CSS import to consumer

farnabaz:
Adds `"sideEffects": false` to all seven packages so bundlers can tree-shake unused exports.

The React, Vue, and Svelte Math components previously imported `katex/dist/katex.min.css` at module level, which blocked this and caused issues in SSR environments. That import has been removed — consumers should now add it once in their project.


atinux (approved):
🚀

PR #234, opened 2026-06-09
fix(comark): treat colon followed by digits as plain text, not a component

antfubot:
### ❓ Type of change

### 📚 Description

A component name must start with a letter (or `$`), but the parser accepted **digit-led** names, with two crash modes:

- **Inline** — `The server is running on :8100` parsed `:8100` as a component:

  ```jsonc
  ["p", {}, "The server is running on ", ["8100", {}]]
  ```

  Renderers then call `createElement('8100')` with a numeric tag name and crash the app.

- **Block** — `:8100` (shorthand) and `::8100` (block) made `parseBlockParams` **throw** `Invalid block params: 8100` during parsing.

**Root cause:** the block name grammar already requires a letter/`$` start (`RE_BLOCK_NAME = /^[a-z$]/i`), but the inline rule (`/[\w$-]/`) and the block pre-filters (`/^:\w/`, and the un-guarded `::` path) accepted digits.

**Fix:** a shared `isComponentNameStart` helper guards all three entry points in `packages/comark/src/plugins/syntax.ts`, mirroring the existing block grammar. A colon sequence whose name doesn't start with a letter or `$` now stays plain text instead of becoming a component or throwing.

Behaviour after the fix:

| Input | Before | After |
|---|---|---|
| `… :8100` (inline) | `['8100', {}]` → `createElement('8100')` crash | text `:8100` |
| `:8100` / `::8100` (block) | throws `Invalid block params` | text `:8100` / `::8100` |
| `:inline-component`, `:badge[New]`, `:h2`, `::alert` | ✅ | ✅ unchanged |

Note: emoji shortcodes like `:100:` are unaffected — the emoji rule runs earlier (`before('emphasis')`) than the inline component rule (`after('entity')`).

Tests live in `packages/comark/test/component-name.test.ts` (9 cases: inline + block regressions plus no-regression checks for valid components). They fail on `main` (2 inline mismatches + 3 block throws) and pass with this change.

### 📝 Checklist

---

This PR was created with the help of an agent.


antfu (approved):
Reviewed.

farnabaz (approved):
Thanks

PR #240, opened 2026-06-11
fix(renderer): escape `[` in markdown text nodes to prevent link re-injection

hendrikheil:
## Problem

When a user writes `\[text\](url)` in a Markdown file (intentionally escaping the brackets so it renders as plain text, not a link), the comark parser correctly strips the backslash escapes and stores a plain text node `"[text](url)"` in the ComarkTree.

However, the markdown renderer returned string nodes **verbatim** — so that text node round-tripped back to `[text](url)`, which the parser then treated as a real link. This created a silent, destructive loop:

1. Save → `[text](url)` (link created where none was intended)
2. Save again → the link's text is now the link label, multiplication of surrounding markers

This was reported in **nuxt-content/nuxt-studio#470** ("MDC autoformatter constantly breaking links and formatting"), where users observed that saving a file containing `\[...\]` patterns would irreversibly convert them to live links and corrupt surrounding bold markers.

## Root cause

`state.ts` `one()` function, string node branch:

```ts
// before
if (typeof node === 'string') {
  if (state.context.html) return escapeHtml(node)
  return node  // ← returned verbatim, no markdown escaping
}
```

## Fix

Escape both `[` and `]` in text nodes when rendering markdown. `[` opens link/image syntax and `]` closes it — a `]` inside link text (e.g. `[foo\] bar](/url)`) would prematurely close the surrounding brackets if left unescaped.

```ts
return escapeMarkdownText(node)

function escapeMarkdownText(text: string): string {
  return text.replace(/[\[\]]/g, (ch) => `\\${ch}`)
}
```

Round-trip stability verified:
- `\[text\] in a sentence.` → AST `"[text] in a sentence."` → renders `\[text\] in a sentence.` → re-parses identically ✓
- `\[Animation Bug\]\(/bugs#section\)` → AST `"[Animation Bug](/bugs#section)"` → renders `\[Animation Bug\](/bugs#section)` → re-parses identically ✓
- `[foo\] bar](/url)` → AST link node with text `"foo] bar"` → renders `[foo\] bar](/url)` → re-parses identically ✓

## Tests

Two updated + one new SPEC file:

- `paragraph-escaped-bracket.md` — basic `\[text\]` round-trip (markdown output updated to reflect `]` escaping)
- `paragraph-link-backslash-escaped.md` — real-world pattern from issue #470 (markdown output updated)
- `link-text-with-closing-bracket.md` — new spec: `]` inside link text must be escaped to avoid premature bracket close


farnabaz (changes requested):
Thanks for the PR @hendrikheil,
Lets also handle `]` chars because they also might create and issue, imagine this example `[dsd\]dsd](/test)` which will become `[dsd]dsd](/test)`  and wrong result

farnabaz (approved):

Issue #145, opened 2026-04-21
Allow-only / exclude specific components


MickL:
Using the `<Comark>` component (and possibly others) it would be handy to have props that allow-only or exclude components.

For example, one may only render p, ul/ol and strong.

atinux:
So what would be the fallback for components not allowed @MickL?

MickL:
Just dont change them, e.g. if strong is excluded keep `**text**`. This is especially important since Comark allows to use components in markdown.

Also a prop to just disallow custom-components, eg`:: callout` would be great!

Feature request on (deprecated) MDC: https://github.com/nuxt-content/mdc/issues/414

atinux:
Would the security plugin with `blockedTags` work in this case for you @MickL to remove `::callout` if not specified?

https://comark.dev/plugins/built-in/security#options-code-blockedtags

atinux:
I think could you start by [making a plugin](https://comark.dev/plugins/custom/plugin-api) to showcase what you have in mind, also happy to list it in community plugins

MickL:
I think it would not work out since every single component needs to be supplied. 

Unfortunately I dont have the time (and the Vercel sponsorship) to work on this project.

atinux:
> I think it would not work out since every single component needs to be supplied.

But so far if you don't give components in the `components` props, it won't try to render the components 🤔

MickL:
In this regard, forgot what I said about components. This issue is only about restricting render of html tags. E.g. imagine a comment system like here on Github, users can add markdown but it should not render a h1:

```
# This should be normal text with, # stays
**This should be bold**
```

farnabaz:
@MickL No markdown renderer do what are you looking for, but due to Comark's design you can achieve this by implementing custom `components-manifest` for Comark component. Here is an example:

- App.vue
```vue
<script setup lang="ts">
import { Comark } from '@comark/vue'

async function componentsResolver(_tag: string) {
  // Check here is create your custom components
  // Allow what you want / render how you want
  // Fallback to empty component if you want to hide it
  return import('./components/TextOnly.vue').then(m => m.default)
}
</script>

<template>
  <Suspense>
    <Comark :components-manifest="componentsResolver" ... />
  </Suspense>
</template>
```

- TextOnly.vue
```vue
<template>
  <span><slot /></span>
</template>
```


If you run this example, you will only see text contents. Then implement components as your prefer.
[...]

PR #245, opened 2026-06-17
feat(plugins): add `allowedTags` and `tagFallback` option into security plugin

maximepvrt:
#145

### ❓ Type of change

### 📝 Checklist


docs/content/4.plugins/1.built-in/security.md
```diff
@@ -165,6 +165,16 @@ security({
...
 
+### `allowedTags`
+
+Tag names to exclusively keep in the AST. Matching is case-insensitive, so, so `SPAN`, `Span`, and `span` are all caught.
+
+```typescript
+security({
+  allowedTags: ['p', 'span', 'ul', 'li', 'ol', 'strong']
+})
+```
```
arashsheyda, line 178: suggestion: should we be mentioning how it would work if user tried with `blockedTags`? which one would be overrides? e.g.

```
blockedTags: ['p'],
allowedTags: ['p'],
```

farnabaz: I think these options should be mutually exclusive, but if user provided both I'd go with `allowedTags` as higher priority.

maximepvrt: Right now, the logic is written to respect both conditions if they are both provided.

I agree that prioritizing `allowedTags` makes the most sense here. The only exception would be if `allowedTags` had a different fallback behavior : like rendering unallowed tags as plain text instead of completely stripping them (e.g., if an end-user who doesn't know the syntax uses brackets [...] in their text, we want to preserve that content as text rather than dropping it entirely just because that specific tag/syntax isn't allowed).

I mentioned this idea in issue https://github.com/comarkdown/comark/issues/145#issuecomment-4736078493

farnabaz: > if an end-user who doesn't know the syntax uses brackets [...] in their text, we want to preserve that content as text rather than dropping it entirely just because that specific tag/syntax isn't allowed

So you want to put back markdown version for unallowed nodes.
I adding markdown renderer here might be not perfect choice for doing so, but plugin can receive an options `replacer: (node) => (node | string | false)` which allows user to replace node according to their desired behavior.

This way, you can call markdown renderer in your custom replacer and preserve markdown version of them, or any custom login (maybe just calling `textContent`)

And default behavior can be the same as current one, if user didn't provide replacer. (I'm 100% sure the `replacer` name though)


WDYT?

maximepvrt: @farnabaz Thanks for the suggestion, I totally agree! I've updated the PR with a custom replacer option `unallowedFallback` with value `false | 'textContent' | 'raw' | ((element: ComarkElement) => any | Promise<any>)`.

Everything is pushed and ready for another look.

docs/content/4.plugins/1.built-in/security.md
```diff
@@ -165,6 +167,36 @@ security({
...
+
+### `unallowedTagsFallback`
+
+Defines the replacement strategy for tags that are filtered out because they are not present in the `allowedTags` whitelist.
+
+::warning
+This option is exclusively effective when used in combination with `allowedTags`. It has no effect on `blockedTags`, which always hard-drops nodes from the AST.
+::
```
farnabaz: Is there any specific reason behind this separation? can't we handle both? I think this is confusing and my first expectation is to handle both in fall back function.

What do you think if we rename:
- `blockedTags -> unsafeTags`
- `allowsTags -> safeTags`
- `unallowedTagsFallback -> unsafeFallback`

Also I think we can change `unsafeFallback` type to `undefined | ((element: ComarkElement) => any | Promise<any>)`
We have simple/on-line utilities which users can use.
```
await renderMarkdown({ nodes: [element] })

textContent(element)
```

maximepvrt: Thanks for the feedback!

My initial thought was to treat `blockedTags` as a strict security measure (like removing `<script>`), while `allowedTags` felt more like formatting choices where a graceful degradation to text/markdown makes more sense than completely deleting the content.

However, your proposal is much simpler and eliminates any confusion. Handling everything through a single fallback function is much more elegant, as users can easily filter and decide whether to return `false`, `textContent(element)`, or `renderMarkdown()`.

maximepvrt: @farnabaz  One last crucial point regarding semantics: using `safeTags` fails to convey the implicit default behavior of a whitelist.

When a developer sees `safeTags: ['p', 'span']`, it looks like a security label, not a strict filter. It doesn't make it obvious that every other tag *not* included in that list will be excluded or stripped by default. On the other hand, `allowedTags` explicitly signals a restriction: if a tag is not 'allowed', it is rejected.

To wrap this up and keep the API clear and intuitive, we could finalize the option names like this:
- `blockedTags`
- `allowedTags`
- `tagFallback`: `undefined | ((element: ComarkElement) => any | Promise<any>)`

farnabaz: @maximepvrt I agree, blocked and allowed sounds good to me 👍

maximepvrt: @farnabaz updated 🚀

farnabaz (approved):
Thanks

farnabaz:
@maximepvrt one last step is required, could you please enable [commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and squash all your commits into a verified commit? Or drop new PR 🙏 

Due to new rule of organization, merges are block and commits must have verified signatures.

maximepvrt:
@farnabaz updated

farnabaz (approved):
Thanks

PR #252, opened 2026-06-25
fix(stringify): render block component children of list items on their own line

hendrikheil:
## Problem

A block component inside a list item (e.g. `::image-split`) was rendered to Markdown glued onto the preceding paragraph line instead of being placed on its own indented line:

````md
- *item one*  ::image-split
  #body
  ### Heading
  ::
````

instead of the expected:

````md
- *item one*
  ::image-split
  #body
  ### Heading
  ::
````

AST and HTML output were already correct — only the Markdown round-trip was broken.

## Root cause

`li.ts` only special-cased a fixed set of block children (`pre`/`blockquote`/`table`/`p`). Component children (rendered via `mdc`) fell through to plain inline concatenation, and the leading spaces came from `mdc` self-applying a fixed level-1 (2-space) indent whenever it has any parent — which only coincidentally matched an unordered list's prefix width.

## Fix

In `li.ts`, recognize component children (tags without a native handler) as block-level when the item has no inline text, render them **without a parent** so `mdc` skips its nesting indentation, and let `li` own the indentation via `prefixWidth`. This is also correct for ordered lists (e.g. `10. ` → 4-space continuation), which the old fixed 2-space assumption got wrong.

### DIFF
--- a/packages/comark/src/internal/stringify/handlers/li.ts
+++ b/packages/comark/src/internal/stringify/handlers/li.ts
@@ -3,7 +3,8 @@ import type { ComarkElement, ComarkNode } from 'comark'
 import { indent } from '../../../utils/index.ts'
 import { comarkAttributes, userBlockAttrs } from '../attributes.ts'
 
-// ol/ul indent themselves via the listIndent context.
+// Block elements that need explicit indentation in list items.
+// Note: ol/ul are handled by their own handlers which manage indentation via listIndent context.
 const blockElements = new Set(['pre', 'blockquote', 'table'])
 
 export async function li(node: ComarkElement, state: State) {
@@ -77,7 +78,7 @@ function escapeLeadingNumberDot(str: string): string {
 
   const len = str.length
   const firstChar = str.charCodeAt(0)
-  if (firstChar < 48 || firstChar > 57) return str
+  if (firstChar < 48 || firstChar > 57) return str // Not a digit
 
   let i = 1
   for (; i < len; i++) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/comark/src/internal/stringify/handlers/li.ts [hunk 1]
```diff
@@ -3,8 +3,7 @@ import type { ComarkElement, ComarkNode } from 'comark'
 import { indent } from '../../../utils/index.ts'
 import { comarkAttributes, userBlockAttrs } from '../attributes.ts'
 
-// Block elements that need explicit indentation in list items.
-// Note: ol/ul are handled by their own handlers which manage indentation via listIndent context.
+// ol/ul indent themselves via the listIndent context.
```
farnabaz:
```suggestion
// Block elements that need explicit indentation in list items.
// Note: ol/ul are handled by their own handlers which manage indentation via listIndent context.
```

packages/comark/src/internal/stringify/handlers/li.ts
```diff
@@ -67,7 +77,7 @@ function escapeLeadingNumberDot(str: string): string {
 
   const len = str.length
   const firstChar = str.charCodeAt(0)
-  if (firstChar < 48 || firstChar > 57) return str // Not a digit
+  if (firstChar < 48 || firstChar > 57) return str
```
farnabaz:
```suggestion
  if (firstChar < 48 || firstChar > 57) return str // Not a digit
```

farnabaz (approved):
Thanks

PR #255, opened 2026-06-26
fix(stringify): render ordered-list `start` as native numbering

hendrikheil:
## Problem

An ordered list with a non-1 `start` round-tripped to the verbose `::ol{…}` block form instead of native markdown numbering:

````md
5. First item        ::ol{start="5"}
6. Second item   →    1. First item
7. Third item         2. Second item
                      ::
````

The same path means a legacy `::ol{start="…"}` wrapper (e.g. one written by an older renderer) never normalizes back to a plain list — it stays an MDC component block. AST and HTML were already correct; only the Markdown round-trip was affected.

## Root cause

`handlers/ol.ts` hardcoded the item counter to `order: 1`, and nothing marked `start` as implicit. So `userBlockAttrs('ol', { start })` treated `start` as a genuine user attr and forced the `::ol{…}` wrapper form, losing the native numbering.

## Fix

- `handlers/ol.ts`: seed the item counter from `start` (`order = start >= 1 ? start : 1`) so native numbering carries it.
- `stringify/attributes.ts`: add `ol: { drop: ['start'] }` to `IMPLICIT_ATTRS`, so `start` is recognized as conveyed by the native syntax and never echoes into a wrapper.

## Result (idempotent)

- `5. 6. 7.` → round-trips as native `5. 6. 7.`
- `::ol{start="3"}\n1. …\n::` → normalizes to native `3. 4. …`
- `::ol{start="undefined"}\n1. …\n::` → heals to clean `1. 2. 3.`
- `::ol{attr="value"}` (a genuine non-`start` attr) → still wraps (unchanged)

## Tests

Two new SPECs (`common-mark/ordered-list-start.md`, `COMARK/ordered-list-start-wrapper-normalizes.md`); full suite **1084 passing**. The #214 attribute specs (`attributes/ordered-list.md`, `attributes/wrapped-ol.md`) stay green.

### DIFF
--- a/packages/comark/SPEC/COMARK/ordered-list-start-wrapper-normalizes.md
+++ b/packages/comark/SPEC/COMARK/ordered-list-start-wrapper-normalizes.md
@@ -1,0 +1,51 @@
+## Input
+
+```md
+::ol{start="3"}
+1. apple
+2. banana
+::
+```
+
+## AST
+
+```json
+{
+  "frontmatter": {},
+  "meta": {},
+  "nodes": [
+    [
+      "ol",
+      {
+        "start": "3"
+      },
+      [
+        "li",
+        {},
+        "apple"
+      ],
+      [
+        "li",
+        {},
+        "banana"
+      ]
+    ]
+  ]
+}
+```
+
+## HTML
+
+```html
+<ol start="3">
+  <li>apple</li>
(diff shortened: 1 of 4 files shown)


farnabaz (approved):
Thanks

PR #257, opened 2026-06-28
fix(svelte): sync core plugins

shyakadavis:
### Summary

Hello; If you follow the docs, and do this:
```ts
import emoji from '@comark/svelte/plugins/emoji'
```
It shall throw an error:
```
Cannot find module '@comark/svelte/plugins/emoji' or its corresponding type declarations.
```

This adds missing `@comark/svelte/plugins/*` re-exports for core Comark plugins.

Svelte packages are built with `svelte-package`, which regenerates `dist`, so dist-only synced plugin files are not included in published packages.

If my understanding of this wrong, or if there's a better way to do this, please let me know.

Thanks.

---
### Before
Screenshot 2026-06-28 at 10 33 31

### Verification

- `pnpm --filter comark build`
- `pnpm --filter @comark/svelte build`
- verified `packages/comark-svelte/dist/plugins` includes the missing core plugin entries

### DIFF
--- a/packages/comark-svelte/package.json
+++ b/packages/comark-svelte/package.json
@@ -54,7 +54,7 @@
   },
   "scripts": {
     "stub": "node ../../scripts/stub.mjs",
-    "build": "svelte-package --input src",
+    "build": "svelte-package --input src && node ../../scripts/sync-plugins.mjs",
     "build:watch": "svelte-package --input src --watch",
     "test": "vitest run",
     "check": "svelte-check --tsconfig ./tsconfig.json",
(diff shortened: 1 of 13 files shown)


farnabaz:
Thanks for the PR and description but the fix actually easier, we just need to add ` && node ../../scripts/sync-plugins.mjs` into build script in package.json

```
    "build": "svelte-package --input src && node ../../scripts/sync-plugins.mjs",

```

shyakadavis:
Missed that. Thanks, @farnabaz  Applied.

farnabaz (approved):

farnabaz:
Thanks for your contribution, @shyakadavis! 🙏

This repository requires all commits to have verified signatures, and the commits here were unsigned. To preserve your work, I created a replacement branch with the same change (rebased onto the current `main`) squashed into a single verified, maintainer-signed commit. Your contribution has been preserved and attributed via a GitHub `Co-authored-by:` trailer.

Merged in #260 — this PR was auto-closed in favor of that one.

PR #259, opened 2026-06-29
fix(render): unwrap a bare `shiki` class on `pre` back to a plain fence

hendrikheil:
Keeps shiki-highlighted code blocks round-tripping to a plain fence instead of leaking a `::pre{.shiki}` block-attribute wrapper.

The leak was introduced in #214, which added the `::pre{attr}` wrapper form together with the `userBlockAttrs` strip guard — that guard never matched the bare `shiki` class.

## The fix

The `::pre{attr}` wrapper form treats any `class` that survives `userBlockAttrs` as a *user* attribute. `userBlockAttrs` strips shiki's injected `class`, but the guard only matched `class="shiki "` **with a trailing space** (the multi-token dual-theme form). **Single-theme shiki emits a bare `class="shiki"`**, which slipped through and forced every highlighted code block to serialize as a `::pre{.shiki}` wrapper — especially visible for code blocks nested in component slots (which also produced a duplicate closing `::`).

Widened the guard from `value.startsWith('shiki ')` to `value.startsWith('shiki')`.

## Tests

`packages/comark/test/plugins/highlight.test.ts`:
- bare `class="shiki"` → plain fence
- multi-token `class="shiki shiki-themes … dark:…"` → plain fence (unchanged)
- highlighted code block nested in a component slot → plain fence, no `::pre`

Full suite green (1079 passed, 2 todo); existing shiki SPEC fixtures unaffected.

### DIFF
--- a/packages/comark/src/internal/stringify/attributes.ts
+++ b/packages/comark/src/internal/stringify/attributes.ts
@@ -92,6 +92,11 @@ export function resolveAttribute(attrs: Record<string, unknown>, renderData: Nod
   return attrs[key]
 }
 
+// Separates the highlighter-injected classes from the user's class within a
+// `pre` class string (e.g. `shiki github-dark . my-class`). Written by the
+// highlight plugin, read here and on re-highlight to recover the user portion.
+export const HIGHLIGHT_CLASS_SEPARATOR = '.'
+
 // Implicit attributes the parser injects per tag — they're conveyed by the
 // native markdown syntax (e.g. `as` becomes `> [!NOTE]`, `task-list-item`
 // is implicit in `- [ ]`) so they should not echo back as user attrs.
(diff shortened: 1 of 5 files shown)


farnabaz:
Thanks for the PR @hendrikheil 
I think the changes are somewhat too much and changing line 131 in attributes.ts would be enough:

```diff
-    if (key === 'class' && tag === 'pre' && typeof value === 'string' && value.startsWith('shiki ')) {
+    if (key === 'class' && tag === 'pre' && typeof value === 'string' && value.startsWith('shiki')) {
```

But as for `code` prop, does it come from Nuxt Content / MDC packages? Comark parser does not add `code` prop anymore.

farnabaz:
We can keep `code` attribute in drop list for now to ease the migration and remove it later.

hendrikheil:
The PR really contains two fixes:
1. strip guard: I can collapse it to startsWith('shiki'), I only used /^shiki(?:\s|$)/ so a genuine user class that happens to start with shiki (e.g. ::pre{.shiki-toggle}) isn't stripped as if it were highlighter output. Cheap insurance, but I'm fine going with the simpler form if you'd rather have that 👍 
2. highlighting idempotency: `highlightCodeBlocks` merges the entire existing pre.class as a user class, so running it twice on the same tree produces `shiki . shiki`, which then leaks back out as `::pre{.shiki}` regardless of the strip guard

hendrikheil:
Actually, I'll just fix the `code` issue in nuxt-studio directly, no need to do it in comark imo. I've also dropped the rehighlighting path. Let's keep this PR focused on the primary issue we're encountering in nuxt-studio right now...

farnabaz (approved):
Thanks

PR #261, opened 2026-06-30
fix: restore three PRs clobbered by stale-tree merge of #245

hendrikheil:
Restores three PRs that were reverted by the stale-tree merge of #245 (which was based on a tree older than all of them):

- **#252** — `fix(stringify): render block component children of list items on their own line`
- **#251** — `docs: fix imports in nuxt rendering`
- **#253** — `docs: use new cache url for code examples`

Each commit restores the original PR's content verbatim.


farnabaz (approved):
Wow, what a mess 😬 
Thanks ❤️

PR #263, opened 2026-07-02
docs(deps): install missing dependency for ogImage rendering

arashsheyda:
### ❓ Type of change

### 📚 Description

currently if you try to access the [ogImage](https://comark.dev/_og/s/o_1deua0.png), it's not working (500) as it's broken. this PR adds the missing dep to fix it

### 📝 Checklist


docs/package.json
```diff
@@ -33,6 +33,7 @@
   },
   "devDependencies": {
     "@nuxt/devtools-kit": "catalog:",
+    "@resvg/resvg-js": "catalog:",
```
arashsheyda, line 37: according to the ogImage docs, this is needed for node env, for vercel edge time we need another dep https://nuxtseo.com/docs/og-image/renderers/satori#installation


are the docs running on edge?

farnabaz: Docs are running on Vercel.
I wonder how we have it working now without this dependency?!

arashsheyda: are you sure it's working? for me when I try to access the ogImage I see 500 errors: https://comark.dev/_og/s/o_dzoei9.png
Screenshot 2026-07-14 at 8 11 21 AM

farnabaz: 😱 So based on Docs we need to add `@resvg/resvg-wasm` for Vercel, right?

arashsheyda: yes (if we're running on vercel edge only tho) vercel node should be fine with `@resvg/resvg-js`

farnabaz (approved):
Thanks ❤️

PR #267, opened 2026-07-04
fix(stringify): escape inline and block markdown syntax in text nodes

### LINKED ISSUE #265: `renderMarkdown` escapes `[` and `]` but not inline and block markdown syntax in text nodes
### Environment

- os: `Ubuntu 24`
- node: `v24.17.0`
- pm: `pnpm@11.9.0`
- framework: `Agnostic`

### Comark Version

v0.5.0

### Reproduction

```ts

const tree = { frontmatter: {}, meta: {}, nodes: [['p', {}, 'use `ls` here']] }

await renderMarkdown(tree)
// actual:   'use `ls` here'
// expected: 'use \\`ls\\` here'

await parse(await renderMarkdown(tree))
// [["p",{},"use ",["code",{},"ls"]," here"]]   ← literal text became a code span

// #240 handles brackets the same way:
await renderMarkdown({ frontmatter: {}, meta: {}, nodes: [['p', {}, 'a [x] b']] })
// 'a \\[x\\] b'
```

### Description

[`escapeMarkdownText`](https://github.com/comarkdown/comark/blob/7e8cd02a1b05b6ef35d741637f0ec957abe66e85/packages/comark/src/internal/stringify/state.ts#L214-L216) only escapes `[` and `]` (added in #240); backticks in a text node round-trip into a code span.

I'll open a PR as soon as I'm able to.

### Additional context

_No response_

### Logs

```shell-script

```

sandros94:
### 🔗 Linked issue

Resolves #265 

### ❓ Type of change

### 📚 Description

Follow-up to #240. Text-node escaping only covered `[`, `]`, so other markdown characters round-tripped into unintended constructs (#265: backticks → code span; also `*`/`_` → emphasis, `<div>` → raw HTML, `&amp`; → entity, leading `#`/`-`/`>` → block constructs).

Now escapes the full inline set (`\` `*` `_` `<` `&` `~` alongside `[` and `]`) wherever it appears, and block markers (`#`, `>`, `-`, `+`, `=`, ordered N.) only at the start of a line — position-aware, so prose like `snake_case`, `a < b`, `C#`, `1.2` mid-line `#` and plain `"use \`ls\` to list dir content` (fun fact, I just realized that even gh one is broken) stays clean. Raw-HTML-block content is left verbatim.

Adds a per-character SPEC fixture for each case plus a unit test covering the negative (no-over-escape) conditions and HTML-block behavior. Closes #265.

### 📝 Checklist

- [x] I have run `pnpm verify` and it passes. (something else was failing before)


sandros94:
While working on fb413b1 I've also noticed that `***` gets -> `---` resulting in double `<hr>`. But I didn't want to include it in this PR (also because I want to double check it first)

farnabaz (approved):
Thanks for the PR

LGTM 👍

PR #269, opened 2026-07-05
fix(comark): only complete unclosed frontmatter while streaming

### LINKED ISSUE #268: bug: auto-close mishandles a leading `---`: doubles a thematic break, and swallows body content
### Environment

- os: `Ubuntu 24`
- node: `v24.17.0`
- pm: `pnpm@11.9.0`
- framework: `Agnostic`

### Comark Version

v0.5.0

### Description

A leading `---` is always treated by `autoCloseMarkdown` as an unclosed frontmatter opener and completed with a fabricated closing `---`. In a complete (non-streaming) document that leading `---` is a thematic break, so the completion is wrong and causes two problems:

1. A lone `---` doubles into two thematic breaks.

```ts
await parse('---')   // [['hr', {}], ['hr', {}]]  — expected one
await parse('----')  // [['hr', {}]]              ✓ (only exact `---` is affected)
```

Auto-close rewrites `---` → `---\n---`; the empty block is rejected by `parseFrontmatter`, so it falls through to the tokenizer as two `---` lines. (This also breaks `hr` round-tripping, since `renderMarkdown` emits every `hr` as `---`.)

2. A leading `---` followed by body silently swallows the body.

```ts
await parse('---\n# Heading')   // frontmatter: null, nodes: []   — the heading disappears
await parse('---\nHello world') // frontmatter: {}, nodes: []     — the paragraph disappears
```

Auto-close rewrites `---\n# Heading` → `---\n# Heading\n---`, which now looks like valid frontmatter, so `parseFrontmatter` consumes the body as YAML and drops it from the output — contradicting its own contract that an unclosed `---` stays content ([test/frontmatter.test.ts → "should handle frontmatter without closing delimiter"](https://github.com/comarkdown/comark/blob/7e8cd02a1b05b6ef35d741637f0ec957abe66e85/packages/comark/test/frontmatter.test.ts#L90-L96)).

Root cause: frontmatter completion only makes sense while streaming (when more text may still arrive). Applied to a finished document, a leading `---` with no closing delimiter is a thematic break, not unfinished frontmatter.

Both reproduce on `main` with default options; `parse(input, { autoClose: false })` returns the correct result in every case.

sandros94:
### 🔗 Linked issue

Resolves #268

### ❓ Type of change

### 📚 Description

`autoCloseMarkdown` treated any leading `---` as an unclosed frontmatter opener and appended a closing `---`. That completion only makes sense while streaming; applied to a complete document it caused two bugs:

- **Doubling**: `---` → `---\n---` → parsed as **two** thematic breaks (and no `hr` round-tripped, since `renderMarkdown` emits `---`).
- **Content loss**: `---\n<body>` → `---\n<body>\n---` → looked like valid frontmatter, so `parseFrontmatter` consumed the body as YAML and **dropped it** — contradicting its own contract that an unclosed `---` stays content.

**Fix**: gate frontmatter completion on `streaming`. `autoCloseMarkdown` gains an optional `{ streaming }` argument (default `false`, backward-compatible — it has no other callers), threaded from `parse` via `opts.streaming`. A complete document now leaves a leading unclosed `---` alone, so the tokenizer reads it as a thematic break and keeps any following body; while streaming, partial frontmatter is still auto-completed exactly as before. A secondary non-empty guard means a lone `---` is never doubled, even mid-stream.

**Behavior**

| input (non-streaming) | before | after |
|-|-|-|
| `---` | `[hr, hr]` | `[hr]` |
| `---\n# Heading` | `[]` (heading lost) | `[hr, h1]` |
| `---\ntitle: x\n---\n\n# Body` | frontmatter + body | frontmatter + body (unchanged) |
| streaming `---\ntitle: x` | completes to frontmatter | completes to frontmatter (unchanged) |

**Tests**

- **SPEC horizontal-rule-bare**: a bare `---` parses to a single `hr`.
- **SPEC frontmatter-unclosed-is-thematic-break**: a leading `---` + body stays a thematic break + body (swallow).
- **test/auto-close.test.ts**: the `frontmatter partial` cases now assert streaming completion, plus new non-streaming cases asserting the body is preserved.

### 📝 Checklist

- [x] I have run `pnpm verify` and it passes. (pre-existing issues)

### DIFF
--- a/packages/comark/SPEC/COMARK/frontmatter-unclosed-is-thematic-break.md
+++ b/packages/comark/SPEC/COMARK/frontmatter-unclosed-is-thematic-break.md
@@ -1,0 +1,43 @@
+## Input
+
+```md
+---
+# Heading
+```
+
+## AST
+
+```json
+{
+  "frontmatter": {},
+  "meta": {},
+  "nodes": [
+    [
+      "hr",
+      {}
+    ],
+    [
+      "h1",
+      {
+        "id": "heading"
+      },
+      "Heading"
+    ]
+  ]
+}
+```
+
+## HTML
+
+```html
+<hr />
+<h1 id="heading">Heading</h1>
+```
+
+## Markdown
+
+```md
+---
+
+# Heading
+```
--- a/packages/comark/SPEC/common-mark/horizontal-rule-bare.md
+++ b/packages/comark/SPEC/common-mark/horizontal-rule-bare.md
@@ -1,0 +1,32 @@
+## Input
+
+```md
+---
+```
+
+## AST
+
+```json
+{
+  "frontmatter": {},
+  "meta": {},
+  "nodes": [
+    [
+      "hr",
+      {}
+    ]
+  ]
+}
+```
+
+## HTML
+
+```html
+<hr />
+```
+
+## Markdown
+
+```md
+---
+```
--- a/packages/comark/src/internal/parse/auto-close/index.ts
+++ b/packages/comark/src/internal/parse/auto-close/index.ts
@@ -62,7 +62,7 @@ export function autoCloseMarkdown(markdown: string, options: { frontmatter?: boo
     }
 
     // Frontmatter: only starts at document line 0
-    if (idx === 0 && trimmed === '---') {
+    if (idx === 0 && options.frontmatter && trimmed === '---') {
       inFrontmatter = true
       continue
     }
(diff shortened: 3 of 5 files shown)


sandros94:
just to be clear, as I stated in the related issue https://github.com/comarkdown/comark/issues/268#issuecomment-4886721864: I'm not really satisfied with this approach, as it basically shifts the issue into another place (essentially the buggy situation happens at streaming rather than static content).

I just wanted to open this PR to hope for inspiration, happy if it gets discarded/closed anyway

farnabaz:
Thanks @sandros94 for the Pr, But I'm thinking why we need to have thematic PR right at the beginning? Having an extra line before this hr will resolve the issue 🤔

sandros94:
> Thanks @sandros94 for the Pr, But I'm thinking why we need to have thematic PR right at the beginning? Having an extra line before this hr will resolve the issue 🤔 

True but this wont solve a broken frontmatter, potentially causing a it to throw.

Also another thing that come to mind is that `renderMarkdown` calls `.trim()` internally which would not solve the issue in node/block-level parsers.

One such use case I'm thinking of is someone pasing a markdown string from clipboard that needs to be parsed into AST before injecting it into an editor. If at any level that string gets trimmed and do happens to start with `---` we are back to square one with this issue

farnabaz (approved):
LGTM 👍
Thanks

PR #272, opened 2026-07-06
fix: lazy-import katex in math plugin helpers

### LINKED ISSUE #271: Math plugin statically imports `katex` despite it being declared an optional peerDependency
### Environment

- Comark Version:   `0.5.0`
- Framework:        `Vue`

### Comark Version

0.5.0

### Reproduction

none

### Description

In `packages/comark/package.json`, katex is declared as an optional peerDependency:

    "peerDependencies": { "katex": "catalog:", ... },
    "peerDependenciesMeta": { "katex": { "optional": true } }

But `packages/comark/src/plugins/math.ts` imports katex statically at the top of the module:

    import katex from 'katex'

Because the import is static and top-level, `import ... from 'comark/plugins/math'` throws `ERR_MODULE_NOT_FOUND` at module load, before any function runs. This makes katex de-facto mandatory, contradicting the optional peerDependency declaration.

katex is only used by the exported `renderMath` and `validateMath` helpers. The plugin itself (the markdown-it rules registered via `markdownItMath`) only emits `math_inline` / `math_block` tokens holding the raw LaTeX in `token.content`, and never calls katex. Those two helpers aren't called anywhere else in the repo either — the framework packages (`@comark/vue`, `@comark/react`, `@comark/svelte`, `@comark/angular`) each render via their own `Math` component that imports katex independently.

Proposed fix: make the katex import lazy — replace the top-level `import katex from 'katex'` with a dynamic `await import('katex')` inside `renderMath` and `validateMath`. This makes those two helpers async (the only breaking change) but decouples katex from module load, so the plugin — which only produces tokens — works with katex absent, consistent with the optional peerDependency. The peerDependency setup itself needs no change.

### Additional context

_No response_

### Logs

```shell-script

```

saschabuehrle:
Fixes #271

katex is declared as an optional peer dep but the static top-level import makes it blow up with ERR_MODULE_NOT_FOUND when katex isn't installed. The plugin itself never calls katex directly — only renderMath and validateMath do.

Switched both helpers to dynamic import('katex') so the module loads fine without katex. Makes both functions async (breaking for direct callers, but neither is called internally).

— saschabuehrle


farnabaz:
Thanks @saschabuehrle
Due to Org rule your PR is blocked, do you mind enabling Commit Signing https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits and update commits in this PR?

farnabaz:
close in favor of https://github.com/comarkdown/comark/pull/350

PR #278, opened 2026-07-10
feat(emoji): add `extend` option for custom shortcodes

benjamincanac:
## Summary

Adds an `extend` option to the emoji plugin so users can add or override shortcodes — including bringing the **complete GitHub set** via their own dependency, without comark bundling a large dataset.

This follows the same "bring your own dataset" pattern as Nuxt UI's [Editor emoji menu](https://ui.nuxt.com/docs/components/editor-emoji-menu), where `@tiptap/extension-emoji`'s `gitHubEmojis` is installed separately and passed in.

## What's in here

- **New `extend` option** — `Record<string, string>` merged on top of the built-in set (values override built-ins, so it also remaps):
  ```ts
  emoji({ extend: { shipit: '🚀', myteam: '🦄' } })
  ```
- **Full set = bring your own** — the built-in set stays small; for the complete GitHub set, install a dataset such as [`gemoji`](https://github.com/wooorm/gemoji) and forward its map:
  ```ts
  import { nameToEmoji } from 'gemoji'
  emoji({ extend: nameToEmoji })
  ```
  No new runtime **or** dev dependency added to comark — the dataset lives in the consumer's project.
- **Bug fixes** in the curated set:
  - `:orange_book:` was a corrupted value (mojibake) → now renders 📙.
  - `:mouse:` (🐭) was silently overridden by a duplicate key for the computer mouse; the animal is restored and the computer mouse moved to `:computer_mouse:` (🖱️). Redundant `zap`/`fire`/`star` duplicates removed.
- **Docs** updated with the `extend` option and the full-set (BYO dataset) pattern.

Backward-compatible: `emoji()` with no options works exactly as before.

## Verification

- ✅ 1,103 `comark` tests pass — emoji SPEC fixtures unchanged.
- ✅ 64 `@comark/vue` tests pass (plugin dedup/idempotency unaffected).
- ✅ Verified end-to-end: base `:rocket:`→🚀, `:orange_book:`→📙, `:mouse:`→🐭, `:computer_mouse:`→🖱️, `:shipit:` via `extend`→🚀, override `:rocket:`→🛸, unknown shortcodes left unchanged.
- ✅ Lint clean (oxlint + oxfmt).

> [!NOTE]
> `pnpm typecheck` reports 3 errors, but all are pre-existing in unrelated test files (`comark-angular/test/rendering.test.ts`, `comark-vue/test/vite.test.ts`) and untouched by this change.

### DIFF
--- a/docs/content/4.plugins/1.built-in/emoji.md
+++ b/docs/content/4.plugins/1.built-in/emoji.md
@@ -18,7 +18,7 @@ links:
     variant: soft
 ---
 
-The `comark/plugins/emoji` plugin converts emoji shortcodes (e.g. `:smile:`) into their corresponding emoji characters. It takes no configuration options.
+The `comark/plugins/emoji` plugin converts emoji shortcodes (e.g. `:smile:`) into their corresponding emoji characters. It ships with a curated set of common emojis, and you can add the full GitHub set or your own shortcodes with the [`extend`](#extend) option.
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


farnabaz (approved):
Thanks

PR #288, opened 2026-07-22
fix: handle bracketed spans inside link labels

### LINKED ISSUE #287: Links with bracketed labels throw "inline rule didn't increment state.pos"
### Environment

- Operating System: `Darwin`
- Node Version:     `v24.12.0`
- Comark Version:   `0.5.1` (also 0.5.0; reproduces on current `main`)
- Framework:        `None` (plain parse(); also hit via @comark/react)

### Comark Version

v0.5.1

### Reproduction

Minimal inline repro (no setup needed):

```typescript

await parse('[[1] Document](#)')
// Uncaught Error: inline rule didn't increment state.pos
```

Any link whose label starts with a bracketed segment triggers it, e.g. `[[1] Document](#)`, `[[link-name] more](https://example.com)`. This is a common shape for LLM-generated citation links, so streaming model output through comark crashes on it.

### Description

Seen an issue where having a link with `[]` inside of it breaks parsing.

The `comark_inline_span` inline rule (registered `before('link')`) returns `true` in silent mode without advancing `state.pos`. markdown-it's inline-rule contract requires any rule returning `true` to advance `state.pos`, silent mode included — `skipToken` enforces it.

Sequence for `[[1] Document](#)`:

1. The `link` rule fires on the outer `[` and calls `parseLinkLabel`.
2. `parseLinkLabel` walks the label; at the inner `[` it calls `skipToken`.
3. `skipToken` runs rules silently; `comark_inline_span` matches `[1]`
   (next char is a space, not `(`/`[`), returns `true`, leaves `pos` untouched.
4. `skipToken` throws `inline rule didn't increment state.pos`.

The same input parses fine in plain markdown-it: `<p><a href="#">[1] Document</a></p>` — the label keeps the literal `[1]`.

There's a secondary aspect: even if the crash is avoided, the span rule eats the inner `[1]` when the link label is re-tokenized, so the link text renders as "1 Document" instead of the literal "[1] Document".

### Additional context

I have a fix ready and will open a PR referencing this issue: return `false` from the silent branch (the pattern the built-in emphasis/strikethrough rules use), plus keep bare `[text]` literal inside link labels (`state.linkLevel > 0`) while still honoring explicit `[text]{attrs}` spans there. Bare spans outside links are unchanged.

### Logs

```shell-script

```

deshiknaves:
### 🔗 Linked issue

Resolves #287

### ❓ Type of change

### 📚 Description

Parsing any link whose label contains a bracketed span throws:

```
Error: inline rule didn't increment state.pos
```

Minimal repro:

```js
await parse('[[1] Document](#)') // rejects
```

This shape is common in LLM-generated citation links (`[[1] Source](url)`), and the same input parses fine in plain markdown-it.

**Cause** — `comark_inline_span` (registered `before('link')`) returned `true` in silent mode without advancing `state.pos`. markdown-it's inline-rule contract requires every rule returning `true` to advance `state.pos`, silent mode included — `skipToken` enforces it. When the `link` rule calls `parseLinkLabel` → `skipToken` on the inner `[1]`, the span rule matched silently, left `pos` untouched, and `skipToken` threw.

**Fix** — two parts:

1. Return `false` in silent mode (same pattern as the built-in `emphasis`/`strikethrough` rules), so `parseLinkLabel`'s own bracket-depth tracking consumes the nested `[...]` and the outer link parses. Just advancing `state.pos` in the silent branch is not enough: `parseLinkLabel` treats a multi-char silent match starting with `[` as a nested link and aborts the whole label.
2. Keep bare `[text]` literal inside link labels. Without this, the span rule ate the inner `[1]` when the label was re-tokenized, rendering link text "1 Document" instead of the plain-markdown "[1] Document". The `link` rule tokenizes label children inside `state.linkLevel++/--`, so when `linkLevel > 0` the span rule now only matches explicit `[text]{attrs}` spans.

Behavior after the fix:

| Input | Result |
|---|---|
| `[[1] Document](#)` | link with literal text `[1] Document` (previously: throw) |
| `[a [x]{.y} b](#)` | explicit span still works inside a link label |
| `[content]{.class}` | unchanged |
| `Hello [World]` | bare span outside links unchanged (matches `compare-parsers` behavior) |

Tests added in `test/parse/inline-span.test.ts` covering all rows above.

### 📝 Checklist

> **Note on `pnpm verify`:** lint and all package test suites pass. The `typecheck` step reports 4 errors that are pre-existing on unmodified `main` (verified by running `tsc --noEmit` on a tree identical to `main`) and don't involve files touched by this PR:
>
> - `packages/comark-angular/test/rendering.test.ts:87` — `'streaming' does not exist in type 'ParseOptions<[]>'`
> - `packages/comark-vue/test/vite.test.ts:22` — vite plugin `config` hook not callable (`ObjectHook` union)
> - `packages/comark-vue/test/vite.test.ts:54` — `'codegenNode' does not exist` on template AST node
> - `packages/comark/test/plugins/dedupe.test.ts:27` — `number | undefined` not assignable to `number`


packages/comark/src/plugins/syntax.ts
```diff
@@ -449,7 +449,14 @@ const markdownItInlineSpan: PluginSimple = (md) => {
...
     if (nextChar === '(' || nextChar === '[') return false
 
-    if (silent) return true
+    // Inside a link label, bare `[text]` stays literal (plain-markdown behavior,
+    // e.g. `[[1] Document](#)`); only an explicit `[text]{attrs}` span matches
+    if (state.linkLevel > 0 && nextChar !== '{') return false
+
+    // Returning `true` without advancing `state.pos` breaks the inline-rule
+    // contract (`skipToken` throws); returning `false` lets `parseLinkLabel`'s
+    // own depth tracking consume nested brackets and the outer link parse
```
farnabaz:
```suggestion
    // Returning `false` lets `parseLinkLabel`'s own depth tracking consume nested brackets and the outer link parse
```

farnabaz (approved):
Thanks

PR #290, opened 2026-07-25
fix: preserve image attributes when a title is present

### LINKED ISSUE #289: Image attributes are dropped when title is provided
### Environment

OS:            Ubuntu 24.04.4 LTS (Linux 7.0.0-28-generic, x86_64)
Node:          v25.6.0
npm:           11.8.0
pnpm:          10.33.0
comark:        reproduces on main (v0.4.0) and published v0.3.2
Framework:     Nuxt 4.4.8 (Vue 3.5.35)

### Comark Version

v0.3.2

### Reproduction

https://github.com/nimonian/comark/blob/b38f0fc/packages/comark/test/img-attributes.test.ts

### Description

```ts
expect(await roundTrip('![alt](img.png "A title"){width="200"}')).toBe(
  '![alt](img.png "A title"){width="200"}',
)
```

is failing because `width="200"` is stripped whenever `"A title"` is provided.

In the renderer source we see that this happens because the branch returns early when `"A title"` is detected, and never appends the collected attributes.

### Additional context

_No response_

### Logs

```shell-script

```

nimonian:
Fixes #289.

The image stringify handler returns early on the title branch and never appends the attribute block, so `alt{width="200"}` round-trips to `alt`, dropping width, class, and every other attribute. The parser already keeps them (the tree carries `width`); only rendering was lossy.

The fix appends the collected attributes on both branches (one line in `packages/comark/src/internal/stringify/handlers/img.ts`).

Two commits, so the regression is visible on its own:

1. a failing test covering title + attribute images (the two controls pass, the two title+attribute cases fail)
2. the one-line fix, turning the test green

Run the test:

```
pnpm --filter comark vitest run test/img-attributes.test.ts
```

Verified against current `main`: the new test is 4/4 green with the fix, and `resolve-attributes` / `props` / `brackets` stay green. Note: `test/index.test.ts` flakes independently of this change (some specs assert a 10ms parse budget that loses under CPU load), so it can fail intermittently on a clean `main` too.


farnabaz (approved):
👍 Thanks

farnabaz:
Thanks @nimonian 
Do you mind enabling Commits sign https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits and replace current commits with signed one? It is an organization organization rule. 
Otherwise I can close this one and do it in another PR

nimonian:
Hi @farnabaz! The commits are now signed - hope it's ok now, let me know if there's any more I can do

atinux (approved):

farnabaz (approved):

PR #298, opened 2026-07-28
fix: ignore props with invalid chars

### LINKED ISSUE #283: Low version browsers such as Chrome 109 may generate errors when setting illegal attribute names
### Environment

Comark Version: 0.5.0
Framework: Vue
Chrome Version: 109

### Comark Version

v0.5.0

### Reproduction

Chrome 109
Open the playground: [https://comark.dev/play/editor?example=all-features](https://comark.dev/play/editor?example=all-features)
Clear the editor and paste the markdown below

```
   :Alert{id="1""}
```

### Description

The output of the large model is unstable and may output characters such as Alert {id="1""} with additional quotation marks, which can result in illegal attribute names and error messages in lower versions

Low version browsers such as Chrome 109 may generate errors when setting illegal attribute names

The 'is not a valid attribute name' in the error message refers to a single double quotation mark character. According to HTML standards, attribute names cannot contain characters such as', ',&lt;,&gt;,/,=, etc., so the browser directly throws a DOMException

### Additional context

*No response*

### Logs

```shell-script

```

farnabaz:
### 🔗 Linked issue

resolves #283

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/comark/src/internal/parse/syntax/props.ts
+++ b/packages/comark/src/internal/parse/syntax/props.ts
@@ -55,7 +55,7 @@ export function searchProps(content: string, index = 0) {
         if (char === '=') {
           index += 1
           props.push([key, searchValue()])
-        } else {
+        } else if (key.match(/^[a-z0-9_\-:]+$/gi)) {
           props.push([key, 'true'])
         }
       }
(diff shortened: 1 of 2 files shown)


rock53770:
```
function isValidAttrName(name) {
  try {
    document.createAttribute(name)
    return true
  } catch {
    return false
  }
}
```
Is it better to use this as each browser version supports different attributes?

farnabaz:
@rock53770 `document` only exists in browser environment but Comark is  agnostic and can be run on any environment.   
Also this PR makes sure that attributes only include alphabet,  numbers, `-` and `_`

PR #304, opened 2026-07-29
fix(stringify): join raw-HTML children without blank lines

### LINKED ISSUE #301: `renderMarkdown` inserts blank lines inside raw-HTML elements, splitting them on reparse
### Environment

- Operating System: `Linux (WSL2)`
- Node Version:     `v24.17.0`
- Comark Version:   `0.5.1` (reproduces on current `main`)
- Framework:        `None`

### Comark Version

v0.5.1

### Reproduction

```ts

const md = '<picture>\n<source srcset="a.avif">\n<img src="a.jpg">\n</picture>'
const rendered = await renderMarkdown(await parse(md))
// '<picture>\n<source srcset="a.avif" />\n\n<img src="a.jpg" />\n</picture>'
//                                        ^^ blank line

console.log(JSON.stringify((await parse(rendered)).nodes))
// [["picture",{…}, ["source",…]], ["img",…]]   ← img split out of the element
```

### Description

The [child-join loop in the `html` handler](https://github.com/comarkdown/comark/blob/060dfa955c7d0467f278580ccc27fa315370ac2a/packages/comark/src/internal/stringify/handlers/html.ts#L74-L92) joins a raw-HTML element's children with `blockSeparator` whenever a child classifies as block (the catch-all arm classifies any unlisted tag as block, so `source`, `figcaption`, `option`… all hit it). A blank line terminates the HTML block per CommonMark, so the rendered markdown can't reparse to the original tree — `parse → renderMarkdown → parse` splits the element.

Anything with ≥2 children where one is not a known inline tag fails: `figure>img+figcaption`, `details>summary+p`, `dl`, `select`, `audio` with two sources, nested `div`s. Single-child elements escape it.

The inline variant is worse: `text <picture><source…></picture> tail` gets the separator too (the loop runs regardless of `oneLiner`), which splits the containing _paragraph_ in two on reparse.

Since the parser never nests markdown blocks inside a `$.html` element (blank-line-separated markdown is parsed _outside_ the element), a blank line inside one is never legitimate in markdown output — children should be joined with `\n`, or nothing at all in inline position.

Fix + tests ready locally, PR incoming

### Additional context

Parse-side relative: #235 / #236 are about the parser _tolerating_ hand-written blank lines inside `<details>`; this one is about the stringifier _introducing_ them. No overlap in touched files.

Filing two more round-trip issues alongside this one, all independent, will cross-link once up

### Logs

```shell-script

```

sandros94:
### 🔗 Linked issue

Resolves #301

### ❓ Type of change

### 📚 Description

In markdown output raw-HTML children are now joined with `\n` (nothing at all in inline position) instead of `blockSeparator` — a blank line would terminate the element on reparse. text/html output keeps the previous separator.

Added round-trip tests for the multi-child cases from the issue, plus single-child and mixed-children controls

P.S.: used Fable to help with branching the changes in their own PR and a final validation of my claims

### 📝 Checklist


packages/comark/src/internal/stringify/handlers/html.ts
```diff
@@ -71,6 +71,9 @@ export async function html(node: ComarkElement, state: State, parent?: ComarkEle
     childrenContent.push(await state.one(child, state, node))
   }
 
+  // A blank line inside a raw-HTML element would terminate it on reparse
+  const childSeparator = state.context.html ? state.context.blockSeparator : oneLiner ? '' : '\n'
```
farnabaz, line 75: I'm confused about this ternary
Could you explain the reason and maybe add parenthesis ?

sandros94: It decides which glue that sticks together all raw html elements, the three cases are basically:

1. `text/html` output → keep `blockSeparator`, behavior untouched there (the separator is purely cosmetic in HTML output)
2. markdown output, inline position (`oneLiner`) → empty string, since even a single `\n` between inline HTML tags comes back as a stray text node in the paragraph on reparse
3. markdown output, block position → `\n`, children keep their own lines but never with a blank line in between (which is what was terminating the HTML block and splitting the element)

On the parenthesis its oxfmt that removes them, which is also technically correct because you migth not want to evaluate the second (`oneLiner ? '' : '\n'`) when the first one was already met.

If I'm not mistaken parenthesis would essentially act like doing (which would force an evaluatuon of the latter case even when not needed):

```diff
-  const childSeparator = state.context.html ? state.context.blockSeparator : oneLiner ? '' : '\n'
+  const markdownSeparator = oneLiner ? '' : '\n'
+  const childSeparator = state.context.html ? state.context.blockSeparator : markdownSeparator
```

sandros94: as quick additional examples:

1. text/html output → keeps `blockSeparator` (unchanged behavior)
    ```ts
    await render(tree, { format: 'text/html' })
    // <picture>
    //   <source srcset="a.avif" />
    //
    //   <img src="a.jpg" />
    // </picture>
    ```
2. markdown output, inline position → '' (a `\n` here would leak a text node into the paragraph)
    ```ts
    await renderMarkdown(await parse('text <picture><source srcset="a.avif"><img src="a.jpg"></picture> tail'))
    // text <picture><source srcset="a.avif" /><img src="a.jpg" /></picture> tail
    ```

3. markdown output, block position → '\n' (own lines, but no blank line — that's what was splitting the element)
    ```ts
    await renderMarkdown(await parse('<picture>\n<source srcset="a.avif">\n<img src="a.jpg">\n</picture>'))
    // <picture>
    // <source srcset="a.avif" />
    // <img src="a.jpg" />
    // </picture>
    ```

farnabaz (approved):
Thanks

PR #305, opened 2026-07-29
fix(stringify): render children of inline components in inline form

### LINKED ISSUE #302: `renderMarkdown` emits block directive fences inside inline `:name[…]` components
### Environment

- Operating System: `Linux (WSL2)`
- Node Version:     `v24.17.0`
- Comark Version:   `0.5.1` (reproduces on current `main`)
- Framework:        `None`

### Comark Version

v0.5.1

### Reproduction

```ts

const tree = { frontmatter: {}, meta: {}, nodes: [['p', {}, 'a ', ['alert', {}, ['item', { x: '1' }]], ' b']] }

const rendered = await renderMarkdown(tree)
// 'a :alert[  :::item{x="1"}\n  :::] b'   ← block fence inside the inline span

console.log(JSON.stringify((await parse(rendered)).nodes))
// [["p",{},"a ",["alert",{},"  :::item","\n",":::"]," b"]]   ← fence as literal text
```

### Description

The `mdc` handler [decides inline-vs-block from the element's own position](https://github.com/comarkdown/comark/blob/060dfa955c7d0467f278580ccc27fa315370ac2a/packages/comark/src/internal/stringify/handlers/mdc.ts#L21-L25) (text siblings / inline parent), but its children [render independently](https://github.com/comarkdown/comark/blob/060dfa955c7d0467f278580ccc27fa315370ac2a/packages/comark/src/internal/stringify/handlers/mdc.ts#L32-L36) — an element child with no handler recurses into `mdc`, picks the block `:::fence` form, and that multi-line output lands inside the parent's `:name[…]` brackets, which the parser can't read back.

The fix is basically to propagate the inline decision down: `a :alert[:item{x="1"}] b` parses back to _exactly_ the input tree, so the inline form is the right target, the children just have to render inline too. The [`span` branch](https://github.com/comarkdown/comark/blob/060dfa955c7d0467f278580ccc27fa315370ac2a/packages/comark/src/internal/stringify/handlers/mdc.ts#L41-L48) already gets this right via `INLINE_HTML_ELEMENTS`, which is why spans with element children round-trip fine today. Demoting the parent to block form instead would not help, a fence mid-paragraph breaks the paragraph just the same

Fix + tests ready locally, PR incoming

### Additional context

Genuinely block-only children (a `ul` inside an inline-positioned component) have no markdown representation either way — out of scope here, they degrade to text instead of fence shrapnel with the fix

### Logs

```shell-script

```

sandros94:
### 🔗 Linked issue

Resolves #302

### ❓ Type of change

### 📚 Description

The `mdc` handler now propagates its inline decision to children via `context.inline`, so an inline `:name[…]` never ends up containing block `:::fences`. Same mechanism the `span` branch was already relying on, just generalized.

P.S.: used Fable to help with branching the changes in their own PR and a final validation of my claims

### 📝 Checklist

### DIFF
--- a/packages/comark/src/internal/stringify/handlers/mdc.ts
+++ b/packages/comark/src/internal/stringify/handlers/mdc.ts
@@ -18,16 +18,19 @@ export async function mdc(node: ComarkElement, state: State, parent?: ComarkElem
   const attributeEntries = Object.entries(attributes)
   const hasObjectAttributes = attributeEntries.some(([, value]) => typeof value === 'object')
 
-  // Component is inline if it has text siblings in parent
-  // or is inside an inline HTML element
+  // Component is inline if it has text siblings in parent,
+  // is inside an inline HTML element or an inline-rendering ancestor
   const hasTextSiblings = parent?.some((child, index) => index > 1 && typeof child === 'string') ?? false
   const insideInlineElement = parent !== undefined && INLINE_HTML_ELEMENTS.has(String(parent[0]))
-  let inline = hasTextSiblings || insideInlineElement
+  let inline = hasTextSiblings || insideInlineElement || state.context.inline === true
 
   // if component has object attributes, it cannot be inline
   if (hasObjectAttributes) {
     inline = false
   }
+
+  // Block fences inside `:name[…]` content cannot be parsed back — keep children inline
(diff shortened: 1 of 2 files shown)


farnabaz (approved):
Thanks

PR #309, opened 2026-07-29
fix: do not include standard html tag names in heading slugs

danielroe:
### ❓ Type of change

### 📚 Description

it would surprise me if we wanted to have the names of html elements in the generated heading ids, but I might be wrong...

### 📝 Checklist


farnabaz (approved):
👍 
Thanks

PR #317, opened 2026-07-31
fix(angular): update build scripts to use angular compiler

### LINKED ISSUE #295: `@comark/angular`: published package has no Angular compiler output, breaks JIT apps (NG0202) and AOT apps
The package is built with plain `tsc`, so the published `dist/` contains nothing from the Angular compiler.

Two problems come from that:

- JIT apps (for example webpack builds that ship `@angular/compiler`): the components compile at runtime, but injection fails with NG0202 because the constructors carry no type information.
- AOT apps (normal Angular CLI): `imports: [ComarkComponent]` does not compile at all, the `.d.ts` files have no component declarations.

So at the moment the package works in neither setup without patching.

Suggested fix: build the package with ng-packagr, like basically every published Angular library. The output then works for AOT and JIT and stays compatible across Angular versions. The `./plugins/*` exports would become secondary entry points, the public API stays exactly the same.

If you would rather not touch the build setup: switching the constructor injection to `inject()` in `ComarkComponent`, `ComarkNodeComponent` and `Binding` already fixes the JIT case with zero build changes. It does not help AOT consumers though.

I'm happy to do a PR for either direction, but wanted your opinion first on which one you prefer.

arashsheyda:
#295

### ❓ Type of change

### 📚 Description

attempt to fix #295

### 📝 Checklist


farnabaz:
I'm waiting for @schplitt to verify the fix 👍 https://github.com/comarkdown/comark/issues/295#issuecomment-5138044475

schplitt:
> I'm waiting for @schplitt to verify the fix 👍 [#295 (comment)](https://github.com/comarkdown/comark/issues/295#issuecomment-5138044475)

@arashsheyda The ci is failing

But when fixed (TS 6 also requires tsconfig changes) the issue is resolved. 👍

farnabaz:
@arashsheyda Do you mind updating this PR and fix CI?

farnabaz (approved):
Thanks

PR #321, opened 2026-07-31
fix(parse): keep `attr="true|false"` as string

### LINKED ISSUE #320: `attr="true"` is typed as a boolean but `attr="false"` (and numbers) stay strings
### Environment

- `comark`: `0.5.1`, `@comark/vue`: `0.5.1`
- Node: 24.x

### Comark Version

0.5.1

## Reproduction (AST level)

```ts

const tree = await parse('::x{flag="true" other="false" count="5"}\n::')
console.log(tree.nodes[0][1])
// → { ':flag': 'true', other: 'false', count: '5' }
//     ^^^^^ only `="true"` was promoted to a `:`-binding; `false`/numbers stayed plain strings
```

## Reproduction (render level, `@comark/vue`)

Rendering the parsed tree through `@comark/vue`'s `ComarkRenderer` into a component with typed props produces:

```
[Vue warn]: Invalid prop: type check failed for prop "flag". Expected Boolean, got String with value "false".
[Vue warn]: Invalid prop: type check failed for prop "count". Expected Number with value 5, got String with value "5".
```

with the `Boolean` prop receiving the truthy string `"false"` (so `false` behaves as `true`), while `flag="true"` on the same component resolves correctly to boolean `true`.

### Description

Comark's inline/block attribute processing promotes a plain attribute whose value is exactly the string `"true"` into a typed binding (`:attr`), which the renderer then resolves to the boolean `true`. It does **not** do the same for `"false"` or numeric strings — those remain plain strings. The result is a surprising, internally inconsistent asymmetry:

- `::x{flag="true"}`  → component prop `flag` is boolean `true`
- `::x{flag="false"}` → component prop `flag` is the **string** `"false"` (which is truthy, and triggers a
  Vue "Invalid prop: type check failed" warning for a `Boolean` prop)
- `::x{count="5"}`    → component prop `count` is the **string** `"5"` (Vue warns for a `Number` prop)

This also contradicts the documented rule that plain `prop="value"` attributes are passed as literal strings — `="true"` is the one value that silently violates it.

## Root cause

`packages/comark/src/internal/parse/token-processor.ts` (dist: `dist/internal/parse/token-processor.js`):

```js
// Handle boolean attributes: {bool} -> {":bool": "true"}
if (handleBoolean && !key.startsWith(':') && !key.startsWith('#') && !key.startsWith('.') && value === 'true') {
  attrs[`:${key}`] = 'true'
  continue
}
```

The `{bool}` shorthand (bare attribute, no value) is normalized to a value of `'true'` earlier, so by the time this check runs it is indistinguishable from an explicit `bool="true"`. The `value === 'true'` guard therefore:

1. captures explicit `key="true"` (turning a would-be literal string into a boolean binding), and
2. has no counterpart for `"false"` (or numbers), producing the asymmetry.

## Expected

Consistent behavior. Either:

- **(a)** Only the bare `{bool}` shorthand should become `:bool` — an explicit `bool="true"` should stay a literal string like every other plain attribute (matching the documented behavior); or
- **(b)** Boolean-valued strings should be handled symmetrically, so `bool="false"` also becomes a typed `:bool` binding resolving to boolean `false`.

Option (a) requires the shorthand to be distinguishable from an explicit value at this point (e.g. a sentinel rather than normalizing `{bool}` to `'true'`). Option (b) is the smaller change and matches what most authors intend when they write `x="false"`.

## Note (separate concern)

Full attribute→prop *type coercion* (coercing every plain attribute to its target component's declared prop type, as `@nuxtjs/mdc` did) is a broader feature and arguably a consumer concern — not part of this report. This report is specifically about the `true`/`false` (and numeric) **inconsistency** in the existing boolean-shorthand handling.

### Logs

```shell-script

```

farnabaz:
### 🔗 Linked issue
resolves #320

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/docs/content/2.syntax/2.components.md
+++ b/docs/content/2.syntax/2.components.md
@@ -304,8 +304,7 @@ Nested components can read the enclosing component's resolved props through the 
 - If the `:prefixed` value is a valid JSON literal (e.g. `5`, `true`, `"foo"`, `{"a":1}`), it's used as-is.
 - Otherwise, the value is treated as a dot-path and resolved against `{ frontmatter, meta, data, props }`.
 - Unknown paths resolve to `undefined`: the prop is passed as `undefined` rather than the raw string.
-- Only props authored with the `:` prefix participate in data binding. Plain `prop="value"` attributes are always passed as literal strings.
-
+- Props authored with the `:` prefix participate in data binding. Plain `prop="value"` attributes are passed as literal strings, except the boolean values `"true"` and `"false"`, which are promoted to typed boolean bindings. Other values, including numeric strings such as `"5"`, remain strings.
 
 ::callout{color="info" icon="i-lucide-replace"}
 Need to interpolate a value directly into text rather than a prop? The [Binding plugin](/plugins/built-in/binding) adds a `{{ path || default }}` inline shorthand that resolves against the same render context.
--- a/docs/content/2.syntax/2.components.md
+++ b/docs/content/2.syntax/2.components.md
@@ -304,7 +304,7 @@ Nested components can read the enclosing component's resolved props through the 
 - If the `:prefixed` value is a valid JSON literal (e.g. `5`, `true`, `"foo"`, `{"a":1}`), it's used as-is.
 - Otherwise, the value is treated as a dot-path and resolved against `{ frontmatter, meta, data, props }`.
 - Unknown paths resolve to `undefined`: the prop is passed as `undefined` rather than the raw string.
-- Props authored with the `:` prefix participate in data binding. Plain `prop="value"` attributes are passed as literal strings, except the boolean values `"true"` and `"false"`, which are promoted to typed boolean bindings. Other values, including numeric strings such as `"5"`, remain strings.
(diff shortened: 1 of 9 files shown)


arashsheyda:
should we be updating this line?

https://github.com/comarkdown/comark/blob/b230f65a75377fa3245b377e0873ff781ea0ceee/docs/content/2.syntax/2.components.md?plain=1#L307

```
- Props authored with the `:` prefix participate in data binding. Plain `prop="value"` attributes are passed as literal strings, except the boolean values `"true"` and `"false"`, which are promoted to typed boolean bindings. Other values, including numeric strings such as `"5"`, remain strings.
```

farnabaz:
Keeping `attr="true"` as string is more logical choice and does not conflict with default behavior. 
I update the PR just to ensures, we don't conflict `{bool}` syntax with `{bool="true"}`.

```
{bool} -> is shortcut for {:bool="true"} -> will convert to boolean
{bool="true"} -> will kept as string
```

adamdehaven:
@farnabaz I'm unsure about the implementation. When I opened the issue I was expecting parity to how `@nuxtjs/mdc` handles booleans and numbers 🤔 does this PR provide the equal experience?

Issue #322, opened 2026-07-31
Double blank line inside a nested component body closes it early


adamdehaven:
### Environment

n/a

### Comark Version

0.5.1

### Reproduction

```js
import { parse } from 'comark'
import { defineComarkPlugin } from 'comark/parse'

// Disable indented-code so 4-space nesting isn't a separate confound (see the indented-code issue).
const noIndentedCode = defineComarkPlugin(() => ({
  name: 'x',
  markdownItPlugins: [(md) => md.disable('code')],
}))

const build = (blanks) => `::page-section
---
full-width: true
---
  ::multi-column
  ---
  columns: 3
  ---

${['A', 'B', 'C', 'D', 'E'].map(x =>
  `    ::container\n    ---\n    display: "flex"\n    ---\n    <svg width="10"></svg>\n    ::`
).join('\n' + '\n'.repeat(blanks))}
  ::
::`

// 1 blank line between siblings -> 5 containers (correct)
// 2 blank lines between siblings -> 1 container  (BUG: siblings after the double blank are dropped)
for (const blanks of [1, 2]) {
  const tree = await parse(build(blanks), { plugins: [noIndentedCode()] })
  // count 'container' nodes in tree.nodes ...
}
```

Observed: 1 blank → 5 containers; 2 blanks → 1 container. Expected: 5 in both cases (2+ blank lines are equivalent to 1 as block separators in Markdown).

### Description

When a block component is nested inside another component, a run of **two or more consecutive blank lines** within the component body prematurely terminates the enclosing component. Every sibling that appears after the blank-line run is dropped from that component's subtree.

Single blank lines between siblings parse correctly; only 2+ consecutive blank lines trigger the early close. This is independent of the indented-code-block issue (no `<pre>` is produced).

## Impact

- Silent content loss in nested components whenever authors use extra vertical spacing between children which is a common, harmless-looking Markdown habit.
- **Divergence from `@nuxtjs/mdc`, which treated 2+ blank lines like a single separator.**

## Expected

Consecutive blank lines inside a component body should act as a single block separator (as in CommonMark), not close the enclosing component.

### Logs

```shell-script

```

farnabaz:
fixed in https://github.com/comarkdown/comark/pull/323

PR #323, opened 2026-08-01
fix(syntax): stop 2+ blank lines from closing nested components early

adamdehaven:
### 🔗 Linked issue

#322 

### ❓ Type of change

### 📚 Description

A run of 2+ consecutive blank lines inside a nested `::component` body silently dropped every sibling after the blank run (or, for `#slotname` slots, emitted a spurious empty paragraph).

Root cause: `comark_block_yaml`, `comark_block_shorthand`, and `comark_block_slots` narrowed `state.lineMax` after matching but never restored it, unlike `comark_block`. Since `skipEmptyLines` is bounded by `state.lineMax`, the second blank line in a run fell through unskipped and broke the tokenizer's indent check.

- `comark_block_yaml` / `comark_block_shorthand`: removed the stray `lineMax` narrowing (neither rule recurses, so it served no purpose).
- `comark_block_slots`: save/restore `lineMax` around its recursive tokenize call, matching the existing pattern in `comark_block`.
- Verified auto-close/streaming interaction still works with the fix

### 📝 Checklist

---

> [!NOTE]
> 2 pre-existing, unrelated failures in `test/inline-component-roundtrip.test.ts` (single-char attribute names, e.g. `x`, get dropped from inline components).

### DIFF
--- a/packages/comark/SPEC/COMARK/component-nested-multiline-separated.md
+++ b/packages/comark/SPEC/COMARK/component-nested-multiline-separated.md
@@ -1,0 +1,213 @@
+## Input
+
+```md
+::page-section
+---
+full-width: true
+---
+  ::multi-column
+  ---
+  columns: 3
+  ---
+
+    ::container
+    ---
+    display: "flex"
+    ---
+    <svg width="10"></svg>
+    ::
+
+
+
+
+    ::container
+    ---
+    display: "flex"
+    ---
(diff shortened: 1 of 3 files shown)


adamdehaven:
@farnabaz can you take a look at this one as well?

farnabaz (approved):
Thanks

Issue #319, opened 2026-07-31
Empty component props block throws `expected a document, but the input is empty`


adamdehaven:
### Environment

- `comark`: `0.5.1`
- `js-yaml`: `5.2.2` (transitive)
- Node: 24.x
- OS: macOS (platform-independent)

### Comark Version

0.5.1

### Reproduction

```ts
import { parse } from 'comark'

// Empty props block (opening `---` immediately followed by closing `---`)
await parse('::page-section\n---\n---\n#title\nHello\n::')
// → throws: YAMLException: expected a document, but the input is empty
```

Blocks that are **not** empty parse correctly, at any nesting/indentation level:

```ts
await parse('::hero\n---\ntitle: x\n---\n#title\nHi\n::')            // OK
await parse('::a\n  ::b\n  ---\n  title: x\n  ---\n  ::\n::')        // OK (nested + indented)
await parse('::hero\n---\n   \n---\ncontent\n::')                    // throws (whitespace-only block)
```

- **Expected:** an empty (or whitespace-only) props block yields a component with no prop (equivalent to omitting the block). No throw.
- **Actual:** `parse()` rejects with `expected a document, but the input is empty`.

### Description

A block component whose `---` fenced YAML props block is **empty** (or contains only whitespace) crashes the parser. Comark passes the empty block to `js-yaml`, which throws `YAMLException: expected a document, but the input is empty`, and the error propagates out of `parse()`, failing the entire document rather than the one empty block.

`@nuxtjs/mdc` (Comark's predecessor) tolerated empty props blocks and simply produced a component with no props. Real-world authored content contains empty props blocks, so this is a migration blocker.

## Root cause

The block-props rule slices the YAML between the fences and calls `parseYaml` unconditionally:

`packages/comark/src/plugins/syntax.ts` — `comark_block_yaml` (dist: `dist/plugins/syntax.js:~245`):

```js
const yaml = state.src.slice(state.bMarks[startLine + 1], state.eMarks[lineEnd - 1])
const data = parseYaml(yaml)   // yaml === '' when the block is empty
```

`packages/comark/src/internal/yaml.ts` — `parseYaml` (dist: `dist/internal/yaml.js`):

```js
import { load, JSON_SCHEMA } from 'js-yaml'

export function parseYaml(content) {
  return load(content, { schema: JSON_SCHEMA })   // js-yaml throws on empty input
}
```

[...]

PR #324, opened 2026-08-01
fix(yaml): resolve empty props blocks instead of throwing

adamdehaven:
### 🔗 Linked issue

https://github.com/comarkdown/comark/issues/319

### ❓ Type of change

### 📚 Description

Empty, whitespace-only, or comment-only YAML props/frontmatter blocks (e.g. `::hero\n---\n---\ncontent\n::`) threw `YAMLException: expected a document, but the input is empty` because `parseYaml()` used js-yaml's `load()`, which throws whenever it finds zero documents.

`parseYaml()` now uses `loadAll()` and branches on `documents.length`: zero documents resolves to `undefined` (treated like an omitted block), one document parses normally, and more than one still throws — preserving `load()`'s original single-document guard. This avoids matching on the exception's message/reason, which would be fragile given the `^5.2.1` caret-pinned js-yaml dependency.

`parseFrontmatter()` falls back to `{}` via `??` (not `||`) so legitimate falsy scalar documents (`0`, `false`, `null`) aren't miscoerced.

No changes needed in `syntax.ts` or `json-render.ts` — both already handle an `undefined`/falsy result.

### 📝 Checklist

### DIFF
--- a/packages/comark/src/internal/yaml.ts
+++ b/packages/comark/src/internal/yaml.ts
@@ -3,10 +3,6 @@ import { dump, JSON_SCHEMA, loadAll, YAMLException, type DumpOptions } from 'js-
 /**
  * Parse YAML content.
  *
- * Uses `loadAll` rather than `load` so that empty, whitespace-only, or comment-only
- * input (which contains no YAML document) resolves to `undefined` — behaving like an
- * omitted block — instead of throwing. Genuine syntax errors still throw during parsing.
- *
  * @param content - The content to parse
  * @returns The parsed data, or `undefined` when the content has no YAML document
  */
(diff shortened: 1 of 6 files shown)


packages/comark/src/internal/yaml.ts
```diff
@@ -1,12 +1,22 @@
...
+import { dump, JSON_SCHEMA, loadAll, YAMLException, type DumpOptions } from 'js-yaml'
 
 /**
- * Parse YAML content
+ * Parse YAML content.
+ *
+ * Uses `loadAll` rather than `load` so that empty, whitespace-only, or comment-only
+ * input (which contains no YAML document) resolves to `undefined` — behaving like an
+ * omitted block — instead of throwing. Genuine syntax errors still throw during parsing.
+ *
```
farnabaz:
```suggestion
```

farnabaz (approved):
Thanks

PR #326, opened 2026-08-04
feat(parse): configuration for default plugins

### LINKED ISSUE #314: Configuring/disabling default plugins
### Description

Currently, it is impossible to configure or disable default plugins (`alert`, `taskList`, `syntax`), as they are completely hardcoded.

https://github.com/comarkdown/comark/blob/0cd5f831d0786e6fbcd55dad7dfecff9cf0cbe24/packages/comark/src/parse.ts#L71

It would be nice if we can disable (or configure `syntax`) these plugins.

### Motivation

I'm currently using `@nuxtjs/mdc` for Markdown parser and renderer, and seeing it is being deprecated with seemingly much better `comark`, I've decided to migrate it. However, this issue is currently blocking my migration.

Previously, I made two versions of Markdown parser: one with all the features like components and everything for contents generated by trusted users, and one with restricted features for contents generated by any users. Obviously you don't want random users inserting any components.

With `@nuxtjs/mdc`, it was really easy to disable any default plugin, including `remark-mdc` which disables the component syntax. But I cannot disable `syntax` plugin in `comark`.

While it is possible to use `markdown-exit` or `markdown-it` for untrusted content and `comark` for trusted content, parsing both types of content with single package and single API, then rendering them with `@comark/vue` or `@comark/nuxt` is much cleaner.

### Proposed solution

Adding three properties in `ParseOptions`:

```ts
interface ParseOptions {
  // ...

  alert?: boolean

  syntax?: boolean | SyntaxOptions,

  taskList?: boolean

  // ...
}
```

and some `if`s in `createParse()`

### Alternatives considered

1. Putting default plugins back in `dedupePlugins`

Something like this:

```js
dedupePlugins([...(options.plugins ? [...options.plugins] : []), alert(), taskList(), syntax()])
```

Maybe it can break parsing? Also, it still cannot disable them (stub plugin may do that but it's not elegant)

2. Adding something like `defaultPlugins` in `ParseOptions` (not flexible)

### Additional context

_No response_

Snack-X:
### 🔗 Linked issue

Resolves: #314

### ❓ Type of change

### 📚 Description

This PR makes default plugins like `alert`, `taskList`, `syntax`, and frontmatter parsing optional (but still defaults to true), making parser can behave more like "vanilla" CommonMark/GFM parser.

Originally, I thought it would be just few options in `ParserOptions` and few `if`s, but LLM-powered review surfaced issues with `autoClose` where syntaxes like `::` still gets auto-closed. PR also deals with those.

Additionally, passing plugin instances with the same name as default plugins (`alert`, `task-list`, `syntax`) will also disable corresponding plugins. This caused small restructuring of syntax plugin to correctly distinguish between "plugin named syntax" and "actual syntax plugin from comark" for configuring auto close.

### 📝 Checklist


Snack-X:
* [x] I haven't updated `bundle.test.ts` yet (running `vitest bundle` reported changes in other packages as well which doesn't look right? I'll update with what CI reports)
* [x] Commits are not signed yet

farnabaz:
Thanks @Snack-X for the PR,
- Instead of having a separate option for each plugins, it would be nice to have `registerDefaultPlugins: boolean (default to true)` option and prevent registering default plugins.
- Also core ensures to remove duplicate plugins, so if user wants to config or override a plugin, can simply register it in the plugins.

This will prevent having many option and provide a simple way to reduce Comark functionality to bare minimum

Snack-X:
@farnabaz Thanks for the reply! Your suggestion makes much more sense.

Branch is rebased upon latest `main`, and three options are merged into one. Also commits are now signed.

farnabaz (approved):
Thanks

PR #330, opened 2026-08-04
fix(angular): render nested custom components

### LINKED ISSUE #294: `@comark/angular`: custom components inside lists, tables or other components render nothing
Hi, thanks for the Angular package, we use it for rendering AI output at work.

We noticed that custom components only work at top level. As soon as one is nested inside a list, a table cell or another component, it gets created but renders empty.

Repro (`badge` is any registered custom component):

```mdc
works: :badge[shown]

- gone: :badge[gone]

::card
:badge[gone]
::
```

The cause is in `comark-node.component.ts`, `renderChildren()`. The nested `ComarkNodeComponent` gets its inputs set by writing directly on the instance:

```ts
componentRef.instance.node = child
```

The component only renders in `ngOnChanges`, and that never fires for direct writes. Changing it to `componentRef.setInput('node', child)` (same for the other inputs) fixes it. We verified this with a patch in our app.

One more thing we found while debugging: there is no try/catch around component creation. When a custom component throws in its constructor, it disappears silently together with all its later siblings, nothing in the console at all. That made this quite hard to track down. A `console.error` with the tag name would already help a lot.

I'm happy to open a PR for both, including a regression test for nested components (inside a list item, a table cell and inside another component).

tomatotomata:
## Summary

Resolves #294.

Nested custom Angular components were initialized by assigning directly to the child component instance. That bypasses Angular input change detection, so nested components were created but rendered empty. This changes the recursive renderer to use `ComponentRef.setInput` for all child inputs.

The same path now catches component creation and initialization errors, logs the component tag, destroys a partially-created view, and continues rendering later siblings.

## Validation

- Added focused regression coverage for nested input binding.
- Added coverage that a failed custom component is logged without stopping later siblings.
- `git diff --check` passes.

The full package test command is left to CI because this checkout cannot materialize the repository's Windows-invalid `packages/comark/SPEC/common-mark/horizontal-rule-3*.md` fixture path.

Let me know what you think.


farnabaz:
Closing in favor of a signed-commit reimplementation: https://github.com/comarkdown/comark/pull/351

Same changes, with @tomatotomata credited via Co-authored-by so the PR can merge under the signed-commit requirement.

PR #333, opened 2026-08-06
feat(parse): add perf option to time the parse pipeline

atinux:
## What

Adds an injectable timing recorder to `ParserOptions` so the parse pipeline can be profiled:

```ts
const parse = createMarkdownParser({ perf, plugins: [highlight()] })
```

When `perf` is provided, these spans are recorded:

- `comark:autoclose` — `autoCloseMarkdown`
- `comark:pre:<name>` — each plugin's `pre` hook (incl. built-ins like `frontmatter`)
- `comark:tokenize` — markdown-exit parse
- `comark:nodes` — token → AST conversion + unwrapping
- `comark:post:<name>` — each plugin's `post` hook (where `highlight`/`toc`/`alert` do their work — usually the actual bottleneck)

## Design

The `ComarkPerf` contract is two functions (`span`, `measure`), **structural** and **universal**:

- No Node-specific APIs (no `diagnostics_channel`) — works in the browser, where comark's streaming rendering runs.
- Structurally compatible with `content.perf` from `comark-content`, so a content app passes one recorder and comark spans appear directly in its debug timelines (see comarkdown/comark-content#99, which consumes this via the pkg.pr.new build until this is released as 0.6.x).
- No-op with zero overhead when omitted (shared frozen no-op, no conditionals at call sites).

## Testing

- `packages/comark/test/perf.test.ts`: span names/order, per-plugin naming for user plugins, identical output with/without `perf`, streaming path.
- Full `comark` package suite passes locally (one pre-existing flaky timeout in `index.test.ts` under load, unrelated; `main` shows similar local flakes).
- Docs: new "Timing the parse" section on the parse API page.


farnabaz (approved):
👍

PR #346, opened 2026-08-11
fix(angular): use inject() instead of constructor DI (fixes NG0202 under AOT)

### LINKED ISSUE #345: Angular: render components throw NG0202 under AOT (constructor DI without Ivy metadata)
### Environment

- `@comark/angular`: `0.6.2` (and `comark` `0.6.2`)
- Angular: `20.3.18`
- Builder: `@angular-devkit/build-angular:application` (esbuild, AOT). Also reproduced in a JIT `TestBed` (Jest + jest-preset-angular).
- TypeScript: `5.x`

### Comark Version

0.6.2

### Reproduction

Any AOT Angular app that renders `<comark-markdown>` (or `<comark-markdown-document>`) triggers it:

```ts
import { Component } from '@angular/core';
import { Markdown } from '@comark/angular';

@Component({
  selector: 'app-root',
  standalone: true,
  imports: [Markdown],
  template: `<comark-markdown [value]="'# Hello **world**'" />`,
})
export class AppComponent {}
```

Rendering throws at instantiation:

```
NG0202: This constructor is not compatible with Angular Dependency Injection
because its dependency at index 0 of the parameter list is invalid.
    at ɵɵinvalidFactoryDep
    at NodeInjectorFactory.MarkdownNode_Factory
```

### Description

**`@comark/angular`'s render components cannot be instantiated under Angular's AOT compiler** (nor under a metadata-less JIT runtime such as `TestBed`).

Root cause: the package is published as **plain compiled output** — no `ng-packagr` partial-Ivy factories (`ɵfac`) and no emitted decorator metadata (`__metadata("design:paramtypes", …)`). Several components declare their dependencies as **constructor parameters**:

```ts
// markdown-node.component.ts — the core recursive renderer
constructor(
  private vcr: ViewContainerRef,
  private renderer: Renderer2,
  private elementRef: ElementRef,
  private injector: Injector
) {}
```

Without an Ivy factory or reflected parameter types, Angular has no way to resolve those parameters, so the injector throws `NG0202`. Because `MarkdownNode` is the leaf renderer of the whole tree, *any* non-trivial document fails — even the ones routed through `MarkdownDocument`.

Interestingly, **`MarkdownDocument` is the one component that works**, precisely because it already uses field-level `inject()`:

```ts
// markdown-document.component.ts
cdr = inject(ChangeDetectorRef); // ✅ resolves without metadata
```

`inject()` resolves from the current injection context and needs no reflected metadata, so it is immune to the packaging gap.

Affected components (all use constructor-parameter DI):

- `Markdown` (`markdown.component.ts`)
- `MarkdownNode` (`markdown-node.component.ts`)
- `Binding` (`binding.component.ts`)
- `Math` (`math.component.ts`)
- `Mermaid` (`mermaid.component.ts`)

### Proposed fix

Migrate those five components to field-level `inject()`, matching `MarkdownDocument`. This resolves `NG0202` under AOT with no behavior change. PR incoming.

### Additional context

Verified against a real Angular 20.3 app: patching the installed `@comark/angular` to use `inject()` turned a failing render (NG0202 on every assistant message) into a fully working one, and a previously-red test suite went from 36/56 to 55/56 passing (the remaining failure is unrelated — the async parser simply needs a microtask flush in that particular test).

lme-agicap:
### What

Migrate the five `@comark/angular` components that still used constructor-parameter DI — `Markdown`, `MarkdownNode`, `Binding`, `Math`, `Mermaid` — to field-level `inject()`, matching `MarkdownDocument` which already did so.

Fixes #345.

### Why

`@comark/angular` is published as plain compiled output — no `ng-packagr` Ivy factories (`ɵfac`) and no emitted decorator metadata. A consumer's AOT compiler (and any metadata-less JIT runtime such as `TestBed`) therefore can't resolve constructor-parameter types, and instantiating those components throws `NG0202` ("dependency at index 0 of the parameter list is invalid"). Since `MarkdownNode` is the leaf renderer of the whole tree, every non-trivial document fails — `MarkdownDocument` is the only component that worked, precisely because it already uses `inject()`.

`inject()` resolves from the injection context and needs no reflected metadata, so this is the minimal, behavior-preserving fix and keeps the package consistent with `MarkdownDocument`.

### Compatibility

No change to the supported Angular range. `inject()` has existed since **Angular 14**, well within the package's `>=17.0.0 <22.0.0` peer range, and is already used by `MarkdownDocument` in this same package. Every touched call is a field initializer (a valid injection context), so nothing calls `inject()` outside a supported context.

### Verification

Reproduced NG0202 in a real Angular 20.3 AOT app and in a Jest `TestBed`. Applying this exact change (constructor DI → `inject()`) makes the render tree instantiate correctly: the failing render started working and a consumer test suite went from 36/56 to 55/56 passing (the remaining failure is unrelated — the async parser just needs a microtask flush in that one test). No public API or template changes.

### DIFF
--- a/packages/comark-angular/src/components/binding.component.ts
+++ b/packages/comark-angular/src/components/binding.component.ts
@@ -6,6 +6,7 @@ import {
   Renderer2,
   OnChanges,
   SimpleChanges,
+  inject,
 } from '@angular/core'
 
 /**
@@ -31,10 +32,8 @@ export class Binding implements OnChanges {
    */
   @Input() defaultValue?: string
 
-  constructor(
-    private elementRef: ElementRef,
-    private renderer: Renderer2
-  ) {}
+  private elementRef = inject(ElementRef)
+  private renderer = inject(Renderer2)
 
   ngOnChanges(_changes: SimpleChanges): void {
     const hostEl = this.elementRef.nativeElement as HTMLElement
--- a/packages/comark-angular/src/components/markdown-node.component.ts
+++ b/packages/comark-angular/src/components/markdown-node.component.ts
@@ -12,6 +12,7 @@ import {
   EnvironmentInjector,
   createComponent,
   reflectComponentType,
+  inject,
 } from '@angular/core'
 import type { ElementNode, Node as MarkdownAstNode, NodeRenderData } from 'comark'
 import { pascalCase, resolveAttributes } from 'comark/utils'
@@ -99,12 +100,10 @@ export class MarkdownNode implements OnChanges {
   /** Parent node (for context like `pre` tag detection) */
   @Input() parent?: MarkdownAstNode
 
-  constructor(
-    private vcr: ViewContainerRef,
-    private renderer: Renderer2,
-    private elementRef: ElementRef,
-    private injector: Injector
-  ) {}
+  private vcr = inject(ViewContainerRef)
+  private renderer = inject(Renderer2)
+  private elementRef = inject(ElementRef)
+  private injector = inject(Injector)
 
   ngOnChanges(_changes: SimpleChanges): void {
     this.render()
--- a/packages/comark-angular/src/components/markdown.component.ts
+++ b/packages/comark-angular/src/components/markdown.component.ts
@@ -6,6 +6,7 @@ import {
   ChangeDetectionStrategy,
   ChangeDetectorRef,
   Type,
+  inject,
 } from '@angular/core'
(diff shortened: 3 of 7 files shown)


farnabaz (approved):
Thanks

lme-agicap:
Thanks for the quick review and merge! 🙏

Do you have a rough idea of when a release including this fix will be published? Just so I know whether to wait for it or keep pointing at the PR build in the meantime.

PR #360, opened 2026-08-17
fix: unable to import some types

will-stone:
### What

Fixes some types not exporting correctly. Everything in `/types.ts` isn't currently accessible without the file extension added.

### Why

Currently it's not possible to import some types, e.g.

```
import type { MarkdownDocument } from 'comark'
```


farnabaz (approved):
Thanks

PR #369, opened 2026-08-19
fix: preserve number, boolean, and null types in YAML block props

### LINKED ISSUE #364: YAML block-props lose number/boolean types (nested values are unaffected)
### Environment

- `comark@0.6.2`
- Reproduces with no plugins (default `components` plugin, on by default)
- Not renderer-specific (happens at parse time, before `@comark/vue`/etc.)

### Comark Version

v0.6.2

### Reproduction

```ts
import { parseMarkdown } from 'comark'

const doc = await parseMarkdown(`::comp
---
label: "hello"
count: 3
enabled: false
nested:
  x: 1
  y: true
---
::
`)

console.log(doc.nodes[0][1])
```

**Actual:**
```json
{
  "label": "hello",
  "count": "3",
  "enabled": "false",
  "nested": { "x": 1, "y": true }
}
```

**Expected:**
```json
{
  "label": "hello",
  "count": 3,
  "enabled": false,
  "nested": { "x": 1, "y": true }
}
```

### Description

Top-level scalar values (numbers, booleans) in a component's YAML block props (`::comp\n---\n...\n---\n`) are coerced to strings in the parsed AST. Nested object/array values under the same block keep their real types, so this looks like an inconsistency rather than intended behavior.

Breaks parity with `@nuxtjs/mdc`, which preserved native YAML types for block props with no special authoring syntax. Any component with typed props (e.g. a `count: number` prop) receives a string instead, producing runtime prop-validation warnings/errors in downstream renderers.

### Root cause

- `plugins/components.js`, the `comark_block_yaml` block rule: `parseYaml()` returns correctly-typed data, but every non-string value is then run through `JSON.stringify()` before being stored as the token's attr:
  ```js
  token.attrSet(key, typeof value === 'string' ? value : JSON.stringify(value))
  ```
- `internal/parse/token-processor.js`, `processAttributes()`: re-parses a stringified value back to its real type only if it looks like `{...}` or `[...]`. There's no equivalent branch for a stringified number or boolean, so `"3"` and `"false"` are never restored.

The `:`-prefix mechanism (`resolveAttributes()`) already restores types correctly for `:`-prefixed keys at render time, so the fix likely belongs in the block-yaml-props path specifically — carrying the real `typeof` from `parseYaml()`'s result through to the AST node, rather than round-tripping through a stringified representation. Pattern-matching the stringified text after the fact (e.g. treating any `"3"`/`"true"`/`"false"` string as typed) would incorrectly coerce an explicitly-quoted YAML string like `count: "3"`, since both forms produce the identical string at that point.

### Logs

```shell-script

```

adamdehaven:
### What

- Add `encodeYamlTypedValue`/`decodeYamlTypedValue` in `internal/yaml.ts`. These functions tag a value with a prefix before it is stored as a token attribute string, then restore the value to its real type later.
- Update `comark_block_yaml` to encode every value from a YAML props block, not only non-string values.
- Update `processAttributes` to decode a tagged value, and skip the existing `{}`/`[]` JSON re-parse heuristic for values already restored this way. This prevents a YAML string that looks like JSON (e.g. `tags: "[1,2,3]"`) from getting parsed a second time.

### Why

A YAML block props value lost its real type. The parser converted every value to a string with `JSON.stringify`. A later step restored objects and arrays, but not numbers, booleans, or null. This caused prop type mismatches in framework components (e.g. Vue prop validation warnings). Fixes #364.

### DIFF
--- a/packages/comark/SPEC/COMARK/component-block-props-types.md
+++ b/packages/comark/SPEC/COMARK/component-block-props-types.md
@@ -29,9 +29,9 @@ Component content
       "component",
       {
         "title": "Hello World",
-        "count": "42",
-        "enabled": "true",
-        "hidden": "false",
+        "count": 42,
+        "enabled": true,
+        "hidden": false,
         "tags": [
           "markdown",
           "docs"
@@ -50,7 +50,7 @@ Component content
 ## HTML
 
 ```html
-<component title="Hello World" count="42" enabled hidden="false" tags="[\"markdown\",\"docs\"]" config="{\"theme\":\"dark\",\"debug\":false}">
+<component title="Hello World" count="42" enabled tags="[\"markdown\",\"docs\"]" config="{\"theme\":\"dark\",\"debug\":false}">
   Component content
 </component>
 ```
@@ -61,7 +61,7 @@ Component content
 ::component
 ```yaml [props]
 title: Hello World
-count: '42'
+count: 42
 enabled: true
 hidden: false
 tags:
--- a/packages/comark/SPEC/COMARK/component-nested-multiline-separated.md
+++ b/packages/comark/SPEC/COMARK/component-nested-multiline-separated.md
@@ -67,12 +67,12 @@ full-width: true
     [
       "page-section",
       {
-        "full-width": "true"
+        "full-width": true
       },
       [
         "multi-column",
         {
-          "columns": "3"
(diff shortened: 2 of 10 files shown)


adamdehaven:
Closing in favor of https://github.com/comarkdown/comark/pull/365

PR #371, opened 2026-08-19
fix(security): keep boolean false props and reject non-string href/src

### LINKED ISSUE #367: `security()` plugin silently strips any prop whose value is `false`
### Environment

- `comark@0.6.2`
- Reproduces with `security()` at default options; not tied to `blockedTags`/ `allowedTags`/URL-prefix config.

### Comark Version

v0.6.2

### Reproduction

```ts
import { parseMarkdown } from 'comark'
import security from 'comark/plugins/security'

const doc = await parseMarkdown(`::comp
---
count: 3
enabled: false
---
::
`, { plugins: [security()] })

console.log(doc.nodes[0][1])
```

**Console output:**
```
[comark/plugins/security] removing unsafe attribute: enabled="false"
```

**Actual:** `{ "count": 3 }`
**Expected:** `{ "count": 3, "enabled": false }`

### Description

`comark/plugins/security`'s attribute validation uses the return value `false` as an internal "reject this attribute" sentinel. A prop whose real value is the boolean `false` is indistinguishable from a rejected one, so it gets silently stripped.

### Root cause

```ts
// internal/props-validation.js
export function validateProp(attribute, value, options = {}) {
  attribute = attribute.toLowerCase()... // normalize
  if (attribute.startsWith('on') || unsafeAttributes.includes(attribute)) return false
  if (attribute === 'href' || attribute === 'xlinkhref') return validateUrl(value, 'link', options)
  if (attribute === 'src') return validateUrl(value, 'image', options)
  return value // <- if `value` is the boolean `false`, this is indistinguishable from the reject above
}

export function validateProps(type, props, options = {}) {
  props = Object.fromEntries(entries.flatMap(([name, value]) => {
    const result = validateProp(name, value, options)
    if (result === false) {  // <- can't tell "rejected" from "value happens to be false"
      return []
    }
    return [[name, result]]
  }))
  return props
}
```

Any component prop with a `false` value is affected, on any tag — not tied to a specific attribute name or to any particular authoring syntax.

### Logs

```shell-script

```

adamdehaven:
### What

Fixes #367: the `security()` plugin used the JS value `false` as an internal "reject this attribute" sentinel, indistinguishable from a prop whose real value is the boolean `false`. Such props were silently stripped. Fixed by introducing a dedicated `REJECTED_PROP` symbol.

Also adds test coverage for #364 (YAML block-props number/boolean/null type loss), already fixed in #365, including an end-to-end `parseMarkdown` -> `resolveAttributes({ parseJson: true })` round trip.

### Why

Widening `validateProp`'s value type to accept non-strings surfaced a related gap: `href`/`src`/`xlinkhref` values that aren't strings (e.g. an array produced by the YAML block-props JSON round-trip) were passed through without URL validation. These are now rejected outright.

### DIFF
--- a/packages/comark/src/internal/props-validation.ts
+++ b/packages/comark/src/internal/props-validation.ts
@@ -1,3 +1,10 @@
+/**
+ * Sentinel returned by {@link validateProp} to signal "drop this attribute".
+ * A dedicated symbol (rather than `false`) keeps rejection distinguishable
+ * from a prop whose real value is the boolean `false` (#367).
+ */
+export const REJECTED_PROP = Symbol('comark:rejected-prop')
+
 export const unsafeTags = ['object']
 
 export const unsafeAttributes = ['srcdoc', 'formaction']
@@ -33,7 +40,11 @@ function rewriteToDefaultOrigin(urlStr: string, defaultOrigin: string): string {
   }
 }
 
-function validateUrl(value: string, mode: 'link' | 'image', options: PropsValidationOptions): string | false {
+function validateUrl(
+  value: string,
+  mode: 'link' | 'image',
+  options: PropsValidationOptions
+): string | typeof REJECTED_PROP {
   const {
     allowedLinkPrefixes = ['*'],
     allowedImagePrefixes = ['*'],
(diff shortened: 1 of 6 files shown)


farnabaz (approved):
Thanks

PR #374, opened 2026-08-21
fix: skip _ inside attributes when only the attributes plugin is on

### LINKED ISSUE #362: Extra underscore appended when only using the attributes plugin
### Environment

- OS: `macOS`
- Node: `v24.19.0`
- Package manager: `npm@11.17.0`
- Comark version: `0.6.2`
- Framework: `None` 

### Comark Version

0.6.2

### Reproduction

Here's a test I ran in the Comark repo to reproduce the issue:

```ts
it("renders attributes when only attributes enabled", async () => {
  const tree = await parseMarkdown(
    'this is a [link with an attribute](https://example.com){target="_blank"}',
    {
      registerDefaultPlugins: false,
      plugins: [attributes()],
    },
  );
  expect(tree.nodes).toEqual([
    [
      "p",
      {},
      "this is a ",
      [
        "a",
        {
          href: "https://example.com",
          target: "_blank",
        },
        "link with an attribute",
      ],
    ],
  ]);
});
```

### Description

When the `registerDefaultPlugins` is set to `false`, the `attributes` plugin is manually enabled, and `{target="_blank"}` is applied to a link in the markdown, a superfluous `_` underscore is appended to the end of the document.

### Additional context

_No response_

### Logs

```shell-script

```

BetterAndBetterII:
## Summary
- With only the attributes plugin enabled, target=_blank no longer auto-closes as italic and appends a leftover underscore.
- Auto-close skips attribute braces whenever attributes is registered, without enabling component fences.

Fixes #362

## Test plan
- [x] vitest default-plugins and auto-close tests
- [x] full comark suite 1427 passed, 2 todo

### DIFF
--- a/packages/comark/src/internal/parse/auto-close/index.ts
+++ b/packages/comark/src/internal/parse/auto-close/index.ts
@@ -13,12 +13,18 @@ import { closeTables } from './table.ts'
  * @param options - `frontmatter` completes an unclosed leading frontmatter block.
  *   `syntax: false` disables Comark component-fence handling (`::` closers and
  *   props braces), for input parsed without the components plugin.
+ *   `attributes: true` still treats `{...}` as an attribute scope (so `_` / `*`
+ *   inside values are not closed) without enabling component fences.
  * @returns The markdown with unclosed syntax closed
  */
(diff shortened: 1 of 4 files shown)


farnabaz (approved):
Thanks 👍

PR #375, opened 2026-08-21
fix(comark): denial-of-service hardening for the parser and stringifier

atinux:
### What

Seven small, independent hardening fixes (one commit each): cap inline-HTML lookahead recursion, clamp codeblock highlight-range expansion, make the default task-list rule linear, make the `escapeInline` `<` check linear, tolerate non-string `class` values in `comarkAttributes`, remove catastrophic backtracking in the rangi comark grammar, and bound json-render spec expansion.

### Why

A security scan found that small attacker-controlled markdown inputs could hang or crash the parsing/rendering process: 10k unclosed `<b>` tags overflowed the call stack, a 20-byte `{1-999999999}` fence info allocated ~2GB, 78KB of plain paragraphs cost ~4s in the task-list rule, a 320KB run of `<a` blocked the event loop for ~44s in `renderMarkdown`, a spaced-out heading line took ~9s in the rangi grammar, and a ~1.4KB json-render fence inflated to 2^24 AST nodes. Each fix now has a regression test built from the reported payload; no legitimate-document behavior changes (verified against the full SPEC suite).

---

### DIFF
--- a/packages/comark/src/internal/parse/token-processor.ts
+++ b/packages/comark/src/internal/parse/token-processor.ts
@@ -771,10 +771,18 @@ export function processInlineTokens(tokens: any[], inHeading: boolean = false): 
   return mergeAdjacentTextNodes(nodes)
 }
 
+// Cap on the html_inline lookahead recursion: each non-void opening tag
+// recurses into the following tokens while searching for its matching close,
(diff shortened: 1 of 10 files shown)


farnabaz (approved):
LGTM 👍

PR #377, opened 2026-08-21
fix(html,ansi): escape plugin fallbacks and strip terminal control characters

atinux:
### What

Escapes the raw source in `@comark/html`'s math and mermaid catch branches (via a shared `escapeHtml` helper, deduplicating the binding plugin's copy), and strips C0/C1/DEL control characters from the document before rendering in `@comark/ansi`.

### Why

A security scan found two output-encoding gaps in the renderer packages: when KaTeX or the mermaid renderer throw on malformed input, the fallback interpolated the author-controlled source into raw HTML (stored XSS even without the raw-HTML plugin); and the ANSI renderer passed ESC/BEL/OSC bytes from untrusted markdown straight to the terminal, enabling forged output and clipboard/title manipulation on permissive terminals. The ANSI sanitization copies the document rather than mutating it, and keeps tab/newline intact.

---


farnabaz (approved):
👍

PR #378, opened 2026-08-21
ci: harden the bundle-snapshot workflow against fork PRs

atinux:
### What

Splits the snapshot update into a tokenless `build` job (`permissions: {}`) that runs the PR's code and a privileged `update` job that only applies the resulting artifact, and makes the report job validate the artifact's `pr-number`/`outcome` and verify the PR head SHA against the workflow run before commenting; comment fences now adapt to backtick runs in PR-controlled text.

### Why

A security scan found that fork PR code executed in a job holding a `contents: write` token (script-injection via `$GITHUB_PATH`/`$GITHUB_ENV` into the token-bearing `github-script` step), and that the fork-controlled artifact could drive bot comments on arbitrary issues or escape its own code fence. The `update` job no longer checks out or executes PR code at all.

---

### DIFF
--- a/.github/workflows/bundle-snapshot.yml
+++ b/.github/workflows/bundle-snapshot.yml
@@ -253,17 +253,19 @@ jobs:
             }
 
   # ---------------------------------------------------------------------------
-  # Update: rebuild, refresh the snapshot and commit it back to the PR branch.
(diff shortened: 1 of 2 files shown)


farnabaz (approved):
Great  👍  🔒

PR #379, opened 2026-08-21
fix(docs): protect the generate-page endpoint from anonymous abuse

atinux:
### What

Adds a same-origin check, a per-IP rate limit (5/min, in-memory sliding window), and server-side body validation (zod: prompt ≤ 1000 chars, `mode` enum) to `/api/generate-page`.

### Why

A security scan found the endpoint was an unauthenticated proxy to the project's AI Gateway: any internet client could run arbitrary prompts through `claude-sonnet-4.6` with up to 8 tool steps, billed to the project, with no quota or length bound (the 1000-char cap was client-side only). The origin check fits because the only caller is the browser playground; the rate limiter is per serverless instance, which is a floor worth having even without shared storage.

---

### DIFF
--- a/docs/server/api/generate-page.post.ts
+++ b/docs/server/api/generate-page.post.ts
@@ -49,6 +49,38 @@ Before generating, fetch the documentation you need:
 
 const FETCH_TIMEOUT_MS = 15_000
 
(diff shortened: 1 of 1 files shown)


farnabaz (approved):
LGTM 👍

PR #383, opened 2026-08-23
docs: bump nuxt example code-explorer pin to current HEAD

dbzx10299:
### What

Updates the `::code-explorer` component to point to HEAD so that the Nuxt framework example is accurate. Previously it was pointing to an old commit using deprecated APIs.


farnabaz (approved):
Thanks

farnabaz:
Thanks for PR, I replaced your commit with signed commit due to organization rules.

PR #385, opened 2026-08-26
fix(comark): run default plugin hooks before user plugin hooks

atinux:
### What

Comark now resolves plugins into one canonical order for MarkdownExit registration and `pre` and `post` hooks: unchanged defaults first, followed by explicit plugins in user-defined order. An explicit plugin replaces a same-name default at its listed position, and duplicate explicit names keep their first instance; tests and public documentation cover these semantics.

### Why

The previous registration and lifecycle orders differed, so a plugin could run in a different position depending on which part of its interface Comark invoked. Using one resolved order makes overrides predictable while ensuring user plugins receive the transformations from defaults they did not replace.

### DIFF
--- a/docs/content/4.plugins/index.md
+++ b/docs/content/4.plugins/index.md
@@ -38,7 +38,9 @@ These plugins are **enabled by default** whenever you call `parseMarkdown()` or 
   ::
 ::
 
+::callout{icon="i-lucide-lightbulb"}
 Default plugins run in the order shown above. Plugins you pass through `plugins` run afterward in the order you provide them. If an explicit plugin has the same `name` as a default plugin, it replaces that default and runs in its explicit position. If you provide the same explicit plugin name more than once, only the first instance runs.
+::
 
 ### Disable default plugins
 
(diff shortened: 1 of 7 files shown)


atinux:
Extended per review discussion (I am an AI agent working with @atinux):

- `refactor`: the hoisted names now derive from the registered `defaultPlugins` array instead of a hardcoded set. With `registerDefaultPlugins: false` the hoisting is a no-op and explicit registration order rules (new test).
- Note on `dedupePlugins([...defaultPlugins, ...userPlugins])`: tested, but it breaks override semantics — dedupe keeps the first occurrence, so the default `alert` instance would win and a user's same-name override would silently lose its hooks. The partition over the already-deduped `plugins` keeps the winning instances and only reorders them.
- `pre` hooks now get the same treatment: among defaults only `frontmatter` has one, and running it first means user `pre` hooks receive the stripped body with `state.frontmatter`/`frontmatterText` populated, instead of raw text a rewriting plugin could corrupt. One partitioned list drives both phases; `markdownItPlugins` keep registration order (user first) for rule precedence.

25 tests pass in `default-plugins.test.ts`; full comark suite green (1508).

farnabaz (approved):

PR #392, opened 2026-08-31
feat(comark): support custom auto-close functions

### LINKED ISSUE #391: Allow a custom auto-close function in the Markdown parser
### Description

Allow `ParserOptions.autoClose` to accept a synchronous function. It should receive the unstable Markdown tail after incremental reuse and run before plugin preprocessing and tokenization.

### Motivation

While optimizing [vue-stream-markdown](https://github.com/jinghaihan/vue-stream-markdown), I found Comark's compact document model and incremental parser architecture excellent. Our benchmarks show several-fold parsing performance improvements over the current mdast-based implementation, so I would like to adopt Comark as the parser while retaining custom completion rules for more controlled handling of incomplete Markdown.

### Proposed solution

Allow `autoClose` to accept a boolean or a synchronous function:

`autoClose?: boolean | ((markdown: string, options: AutoCloseOptions) => string)`

`true` would keep the current behavior, `false` would disable auto-close, and a function would replace the built-in implementation at the same stage.

### Alternatives considered

Preprocessing before parsing handles the full source and may disrupt incremental reuse. A `pre` plugin can work, but it is a general transform hook whose execution order may not match the built-in auto-close stage.

### Additional context

PR #385 makes a `pre` plugin less reliable as an auto-close replacement because default hooks run first. A function would provide a stable extension point, and I would be happy to submit a PR if this approach is accepted.

jinghaihan:
### What

Allow `ParserOptions.autoClose` to accept a custom function. In streaming mode, it receives only the unstable tail.

### Why

This enables custom completion rules while preserving Comark’s incremental parsing. Closes #391.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -517,7 +517,7 @@ describe('functionUnderTest', () => {
 ```typescript
 const result = await parseMarkdown(markdownContent, {
   autoUnwrap: true,             // Remove <p> wrappers from single-paragraph containers
-  autoClose: true,              // Auto-close incomplete syntax
+  autoClose: true,              // Auto-close incomplete syntax; also accepts (markdown) => string
   unwrap: 'p',                  // Strip top-level wrapper tags (MDC unwrap); merges paragraphs
   registerDefaultPlugins: true, // frontmatter, html, alert, task-list, components, attributes; false to disable
 })
--- a/docs/content/5.reference/1.parse.md
+++ b/docs/content/5.reference/1.parse.md
@@ -360,7 +360,7 @@ Both `parseMarkdown()` and `createMarkdownParser()` accept the same `ParserOptio
 | Option | Type | Default | Description |
 |--------|------|---------|-------------|
 | `autoUnwrap` | `boolean` | `true` | Remove unnecessary `<p>` wrappers from single-element containers |
-| `autoClose` | `boolean` | `true` | Auto-close incomplete markdown syntax |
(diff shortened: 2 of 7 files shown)


farnabaz:
Thanks for the proposal and PR @jinghaihan. I'm glad that you found Comark suitable to adopt in vue-stream-markdown.
You proposal and changes looks good to me 🙂 

Just small thing before merge, would you rewrite your commits and create new signed commit? https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

This is due to organization rules and currently PR is blocked because of this. If you don't want to go through commit signing setup tell me, I'll rewrite your commit add you as co-author and force push 🙂

jinghaihan:
> Thanks for the proposal and PR @jinghaihan. I'm glad that you found Comark suitable to adopt in vue-stream-markdown. You proposal and changes looks good to me 🙂
> 
> Just small thing before merge, would you rewrite your commits and create new signed commit? [docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)
> 
> This is due to organization rules and currently PR is blocked because of this. If you don't want to go through commit signing setup tell me, I'll rewrite your commit add you as co-author and force push 🙂

Done — I’ve rewritten and force-pushed the commit with a verified SSH signature. Thanks!

farnabaz (approved):
Thanks 👍

PR #393, opened 2026-09-02
fix(shiki): handle array class values when highlighting code

dbzx10299:
### What

When using Shiki with certain transformers, such as the [`transformerNotationDiff`](https://shiki.style/packages/transformers#transformernotationdiff), `child[1].class` is an array with this shape:

```js
[ 'line', 'diff', 'remove' ]
```

If we have a code block:

````
```ts {1}
const foo = 'foo' // [!code ++]
```
````

We will have this html output:

```html
<span class="line,diff,add highlight"></span>
```


farnabaz (approved):
Thanks

PR #408, opened 2026-09-10
fix(vue): accept a parsed document in a defined component

benjamincanac:
### What

`defineMarkdownComponent` and `defineMarkdownDocumentComponent` now reuse the prop tables exported by `<Markdown>` and `<MarkdownDocument>` instead of hand-copying them. That fixes the reported `Invalid prop: type check failed` warning when passing an already parsed document, plus two forwarding gaps, and makes the drift structurally impossible.

> [!NOTE]
> Based on #407 so the parser rebuild is a one-line `computed` rather than an expensive rebuild on every prop change. Review that one first.

### Why

Reported from ui.nuxt.com: `<Markdown>` accepts `value: [String, Object]` and short-circuits on `isMarkdownDocument`, but the component `defineMarkdownComponent` returns declared `value: { type: String }`, so passing a parsed document warned even though it rendered correctly. The underlying cause is that the wrapper carried its own copy of the prop table, and the two had drifted three times. Fixing just `value` would leave the next drift to be found the same way.

#### One prop table

`markdownProps` and `markdownDocumentProps` are exported from the components and the wrappers use them directly. The render calls spread `props` rather than listing each one.

A React-style spread alone would not have been enough here: in Vue, spreading `props` only spreads *declared* props, so `data` and `documentKey` would still have been missed. Sharing the declaration is what actually fixes it.

#### The three gaps this closes

- `value` inherits `[String, Object]`, so a parsed document no longer warns.
- `data` is declared rather than relying on attribute fallthrough.
- `documentKey` is declared on `<Markdown>` and forwarded to `<MarkdownDocument>`, which has been waiting for it. It is what the `globalThis.comarkContext` live-update subscription reads, and it previously needed two fallthrough hops through a component that sets `class` explicitly.

Worth noting for review: `data` did already arrive via fallthrough, so only the `value` warning and `documentKey` were visibly broken. I verified that by reverting the change and watching exactly those two tests fail. `data` is pinned anyway, since fallthrough is not something to depend on.

#### `Exclude` was a no-op

`Exclude<ParserOptions, 'plugins'>` appeared in six places. `Exclude` filters union members, so on an object type it resolves to plain `ParserOptions`. `Omit` is what was meant, and it surfaces a real bug: `:options="{ plugins: [x] }"` type-checked and was then silently discarded, because the component passes `plugins: props.plugins` separately.

This is a type-level breaking change for anyone who was relying on the accident, but it only rejects code that never worked at runtime.

### DIFF
--- a/packages/comark-angular/src/components/markdown.component.ts
+++ b/packages/comark-angular/src/components/markdown.component.ts
@@ -44,7 +44,7 @@ export class Markdown implements OnChanges {
   @Input() value?: string | MarkdownDocumentType
 
   /** Parser options (excluding plugins) */
-  @Input() options: Exclude<ParserOptions, 'plugins'> = {}
+  @Input() options: Omit<ParserOptions, 'plugins'> = {}
 
   /** Additional plugins to use */
   @Input() plugins: ParserOptions['plugins'] = []
--- a/packages/comark-react/src/components/Markdown.tsx
+++ b/packages/comark-react/src/components/Markdown.tsx
@@ -20,7 +20,7 @@ export interface MarkdownProps {
   /**
    * Parser options (excluding plugins)
    */
-  options?: Exclude<ParserOptions, 'plugins'>
+  options?: Omit<ParserOptions, 'plugins'>
 
   /**
    * Parser to use instead of one resolved from `options` and `plugins`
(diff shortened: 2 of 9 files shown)


farnabaz (approved):
LGTM
Thanks

PR #412, opened 2026-09-11
fix(comark): surface parse failures instead of rendering an empty document

benjamincanac:
### What

`createSerializedTask` swallowed every rejection, so a failed parse resolved to `null` and the component rendered an empty `<div class="comark-content">` with nothing in the console. It now keeps the queue alive after a failure but rejects to the caller, and the Vue, Angular and Svelte components that had a bare `.then` handle it: the initial parse may reject so `Suspense` or `onErrorCaptured` sees it, later parses log and keep the last good document. Svelte also advances its version guard on rejection so an older in flight parse cannot overwrite the document after a newer one fails.

### Why

Split out of #407 at review. The helper change alone turns a silent empty render into an unhandled rejection at every call site that never expected one, so each site needed a policy, and the one used is what React already does by letting a rejection reach the nearest boundary.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Markdown error handling across Angular, Svelte, and Vue integrations.
  - Parsing failures now reach the appropriate error-handling paths.
  - Angular preserves the previously rendered document when parsing fails.
  - Svelte prevents stale parsing results from replacing current content after a newer parse fails.
  - Serialized processing continues after failed tasks while preserving each task’s result or error.
- **Tests**
  - Added coverage for parsing failures, stale results, and continued serialized processing.

### DIFF
--- a/packages/comark-angular/src/components/markdown.component.ts
+++ b/packages/comark-angular/src/components/markdown.component.ts
@@ -111,9 +111,12 @@ export class Markdown implements OnChanges {
     }
     source = source.trim()
 
-    this.serializedParse(source, { streaming: this.streaming }).then((result) => {
-      this.document = result
-      this.cdr.markForCheck()
-    })
+    this.serializedParse(source, { streaming: this.streaming })
+      .then((result) => {
+        this.document = result
+        this.cdr.markForCheck()
+      })
+      // Keep the last good document rendered and report the failure.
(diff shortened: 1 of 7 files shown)


farnabaz:
I'm closing this because this issue is fixed on other PR. Feel free to drop another if issue persist on main or something is missing.

PR #417, opened 2026-09-14
fix(components): parse consecutive inline components without spaces

### LINKED ISSUE #256: Consecutive inline elements are treated as text
### Environment

- https://comark.dev/play/editor
- 2026-06-26
- 0.4.0 is the latest released version, so presumably that

### Comark Version

https://comark.dev/play/editor on 2026-06-26

### Reproduction

https://comark.dev/play/editor?example=all-features

with the following content:

```md
:b[text]:i[text]
```

### Description

When two inline elements after one another occur, without a space, the latter is treated as text. Its body is then treated as a span.

#### Input
`:b[text]:i[text]`

#### Actual output
`<b>text</b>:i<span>text</span>`

#### Expected output
`<b>text</b><i>text</i>`

### Additional context

_No response_

### Logs

```shell-script

```

a1stok:
### What

allow inline components to be parsed immediately after other inline elements. added `]`, `}`, `)`, and `>` to the allowed previous characters list in the parser.

### Why

when you wrote `:b[text]:i[text]` the second component wasn't parsed because the `]` before it wasn't whitelisted. this allows components to follow right after other components, attributes, links, and html tags without needing a space. resolves #256

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved inline component parsing when components appear immediately after brackets, braces, parentheses, or angle brackets.
  * Consecutive inline components and components following attributes now render and parse correctly.
  * Improved text escaping to preserve literal colon-prefixed text and prevent unintended component parsing.


farnabaz (approved):
LGTM, thanks

c-harding:
https://github.com/comarkdown/comark/pull/417#discussion_r4063027434

@farnabaz This comment was never resolved. It’s only an issue with a test, but the test would pass with unsafe content.

Issue #296, opened 2026-07-28
Proposal: modernize `@comark/angular` (signal inputs, `inject()`)


schplitt:
The Angular package currently uses `@Input()` decorators and constructor injection everywhere. I would like to do a modernization pass: signal inputs via `input()`, `inject()` instead of constructors, and `computed()` where it makes sense.

There is one catch: signal inputs are stable since Angular 19, and the package currently allows `>=17`. So this would raise the minimum supported Angular version. That is your call as maintainers, which is why I'm asking here first instead of just sending a PR.

For consumers nothing changes. Props on custom components are already set via `setInput()`, and that works the same with signal inputs.

If you're open to this, let me know which Angular version you're comfortable with as the new minimum and I'll prepare the PR. I would keep it separate from the bug fixes (#294, #295) so those can land independently.

atinux:
What do you think @arashsheyda ?

arashsheyda:
I think it would be a good idea to change the minimum to v19 🚀

atinux:
Could we close this issue too as https://github.com/comarkdown/comark/pull/346 is merged?

schplitt:
> Could we close this issue too as https://github.com/comarkdown/comark/pull/346 is merged?

I wouldn't say so to be honest.
A lot of modern angular features are not used currently.

I would be very in favor of it (though this would raise the minimal version for Angular to 19).

I'm unsure if you would be in favor of that or not.
If so let me know and I will look into it myself.

arashsheyda:
@schplitt do you want to open a PR so we can review?

adrian-ub:
Hey there! I started working on this

**PR: [#446](https://github.com/comarkdown/comark/pull/446) — migrate to signal inputs and ng-packagr**

It covers everything proposed here:

- **Signal inputs** — `@Input()` → `input()` across `Markdown`, `MarkdownDocument`, `MarkdownNode`, `Binding`, `Math` and `Mermaid` components
- **`inject()`** instead of constructor injection, plus **`computed()`** where it made sense
- **Build migration to `ng-packagr`** — the hand-rolled `ngc` pipeline is gone; component plugins are now secondary entry points (`plugins/binding`, `plugins/math`, `plugins/mermaid`)
- **Minimum Angular raised to v19**, as agreed in this thread (peer deps now `>=19 <23`)

For consumers nothing changes: props on custom components are still set via `setInput()`.

Happy to adjust anything — would love a review!

[...]

PR #446, opened 2026-09-17
feat(angular)!: migrate to signal inputs and ng-packagr

adrian-ub:
### What

Modernizes `@comark/angular` per #296: components now use signal inputs (`input()`), `inject()` instead of constructor injection, and `computed()` where it makes sense. The build also moves from the hand-rolled `ngc` pipeline to `ng-packagr`, with the component plugins restructured as secondary entry points (`plugins/binding`, `plugins/math`, `plugins/mermaid`).

### Why

The package still relied on `@Input()` decorators and constructor injection. Signal inputs are the stable, modern reactivity model in Angular, and `ng-packagr` is the standard library build tool — it produces correct partial-compilation output for both JIT and AOT consumers with far less custom tooling. As proposed in the issue, this raises the minimum supported Angular version to 19 (peer range now `>=19 <23`). For consumers of the components themselves nothing changes: props are still set via `setInput()`. The only API change is the plugin imports, documented below.

### Breaking change: named plugin imports

The plugin entry points now use **named exports** instead of default exports — `ng-packagr` strictly enforces the Angular Package Format and strips `export default` from entry points. Update your imports:

```diff
- import binding, { Binding, If } from '@comark/angular/plugins/binding'
+ import { binding, Binding, If } from '@comark/angular/plugins/binding'

- import math, { Math } from '@comark/angular/plugins/math'
+ import { math, Math } from '@comark/angular/plugins/math'

- import mermaid, { Mermaid } from '@comark/angular/plugins/mermaid'
+ import { mermaid, Mermaid } from '@comark/angular/plugins/mermaid'
```

The component APIs (Markdown, MarkdownDocument, defineMarkdownComponent) are unchanged.

### Why named exports?

This is *by design* of the Angular Package Format (APF), which Angular's tooling enforces strictly: `ng-packagr` intentionally ignores and removes `export default` from library entry points. The Angular team discourages default exports in libraries for three technical reasons:

- **Static analysis & tree-shaking.** Named exports are predictable — bundlers know exactly which pieces of code an application consumes, which enables dead-code elimination. A default export is an opaque object that forces  bundlers to include the whole module just in case.
- **The Angular compiler (Ivy/ngtsc).** Angular needs strongly typed static references to resolve the metadata of its decorators (`@Component`,  `@NgModule`, `@Injectable`). A default export can't be traced back to a  concrete declaration.
- **Module interoperability.** Default exports historically cause friction when mixing CommonJS and ESM — the classic "nested `.default` property" bug.

## Summary by CodeRabbit

* **New Features**
  * Updated Angular integration for compatibility with Angular 19–22.
  * Added named entry points for the binding, Mermaid, and math plugins.
  * Updated Angular components to respond reactively to input changes while preserving existing rendering behavior.

* **Documentation**
  * Updated Angular examples and guidance to use named plugin exports.

* **Chores**
  * Updated Angular package distribution, build configuration, and test workflows.


farnabaz:
WDYT @arashsheyda ?

PR #450, opened 2026-09-20
fix(props): class props should support tailwind custom values

dbzx10299:
### What

In Comark when using attributes, it's common a user wants to add a TailwindCSS class:
```md
this is some markdown{.text-gray-500)
```

Which renders:

```html
<p class="text-gray-500">This is some markdown</p>
```

But, this does not work with a Tailwind custom value:

```md
this is some markdown{.text-[#7d7d7d]}
```

This renders:

```html
<p class="text-[" id="7d7d7d]">This is some markdown</p>
```
### Why

This fixes the bug with how a hex color value inside a set of brackets `[]` is treated as an `id` prop. It checks inside a class branch for brackets supporting Tailwind custom values.

```md
some markdown text{.text-[#7d7d7d]}
``` 

Output:
```html
<p class="text-[#7d7d7d]">some markdown text</p>
```

It also introduces a flag `bracketAware` so only class props are affected.

AGENTS:
- The What and Why should each be 1-3 sentences.
- Write your answers based off of your conversation with the user, not purely based on the changed code
  - It's helpful to understand the reasoning behind an approach, what alternatives were considered, and why this approach was ultimately selected, rather than a summary of the lines changed
- For complex changes (>1k lines, not counting lockfiles):
  - Add a `### Walkthrough` section to the end which breaks down the primary changes into easy-to-follow `#### Sub-Sections`
-->

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved parsing of class values containing nested or balanced brackets.
  - Prevented delimiters inside bracketed expressions from prematurely ending class values.
  - Preserved class content when bracketed expressions are incomplete.
  - Improved support for custom-value classes, including colors, decimal values, multiple bracketed values, and trailing IDs.
  - Enhanced class property parsing for nested expressions.

- **Documentation**
  - Added examples for Tailwind arbitrary-value classes in component, inline component, and attribute syntax.


farnabaz:
Thanks for the PR @dbzx10299,
Do you mind adding some tests and check review form  Code Rabbit?

dbzx10299:
@farnabaz I added some tests and checked out the review. I have a local example working but there seems to be a lot of edge cases and it's making the code a lot harder to read. If a user writes a wrong syntax like `{.class[broken}...` it is a user error so I don't want to push my local solution because it sacrifices readability and I am not even sure I have a good solution. The Tailwind issue is resolved but I am having a bit of a hard time on this one.

farnabaz:
@dbzx10299 I pushed a small fix to your PR to make sure bracket search do not go outside of `{}` boundary.

farnabaz (approved):
Thanks

dbzx10299:
@farnabaz That's the part I was struggling with and I honestly don't want to push up ai slop. Thanks for the help.

